PR #18, opened 2019-10-20
Add support for packed attestation format (surrogate attestation type)

adrienperonnet:
This PR adds support for surrogate attestation type for packed attestation format and solve https://github.com/Firstyear/webauthn-rs/issues/17.

After that change, I was able to use the example web server to register and authenticate with touchID using Chrome (Version 77.0.3865.120).

- [x] cargo fmt has been run

### DIFF
--- a/src/attestation.rs
+++ b/src/attestation.rs
@@ -80,7 +80,7 @@ pub(crate) fn verify_packed_attestation(
     client_data_hash: &Vec<u8>,
     counter: u32,
 ) -> Result<AttestationType, WebauthnError> {
-    //Verify that attStmt is valid CBOR conforming to the syntax defined above
+    // Verify that attStmt is valid CBOR conforming to the syntax defined above
     // and perform CBOR decoding on it to extract the contained fields
     let att_stmt_map = att_stmt
         .as_object()
--- a/src/attestation.rs
+++ b/src/attestation.rs
@@ -83,14 +83,12 @@ pub(crate) fn verify_packed_attestation(
         (None, None) => {
             let credential_public_key = crypto::COSEKey::try_from(&acd.credential_pk)?;
 
-            //TODO: Validate that alg matches the algorithm of the credentialPublicKey in authenticatorData.
+            //Validate that alg matches the algorithm of the credentialPublicKey in authenticatorData
             let alg = att_stmt_map
                 .get(&serde_cbor::ObjectKey::String("alg".to_string()))
-                .ok_or(WebauthnError::AttestationStatementSigMissing)?;
-            if alg.as_i64() != None {
-                //algorithm -7 ("ES256"),
-                println!("{:?} != {:?}", alg, credential_public_key.key);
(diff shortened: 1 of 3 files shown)


Firstyear:
Thanks for the PR - really appreciate you putting in the time to add this. For the most part it looks really good, I'm about to add a few minor comments inline though. Nothing major, just stylistic choices basically. Thanks again!

src/attestation.rs
```diff
@@ -55,6 +59,74 @@ pub enum AttestationType {
...
+    alg_from_auhenticator_data: &COSEContentType,
+) -> bool {
+    att_stmt_map
+        .get(&serde_cbor::ObjectKey::String("alg".to_string()))
+        .and_then(|alg| alg.as_i64())
+        .map(|alg| alg == i64::from(alg_from_auhenticator_data))
+        .unwrap_or(false)
+}
+
+// https://w3c.github.io/webauthn/#sctn-packed-attestation
```
Firstyear: It's worth taking the text from this section and turning them into comments to inline with the code, so that we can assert all the steps have been taken as per the specification. See verify_fidou2f_attestation for an example of how I did this.

src/lib.rs
```diff
@@ -887,6 +897,32 @@ mod tests {
...
+
+        let chal = Challenge(
+            base64::decode_mode(
+                "lP6mWNAtG+/Vv15iM7lb/XRkdWMvVQ+lTyKwZuOg1Vo=",
+                base64::Base64Mode::Standard,
+            )
+            .unwrap(),
+        );
+
+        // Generated using Touch ID via Chrome
```
Firstyear: Can you be more specific about the platform here? Touch ID on macos via chrome? Or Touch ID on a different version?

src/attestation.rs
adrienperonnet: Good point, I'll add the verification steps from the spec into inline comment.

src/lib.rs
adrienperonnet: I've used Touch ID on MacOS via Chrome, I'll add those details in the comment.

Firstyear: I'm really curious about how you used touchid on chrome. Do you mind sending me some instructions? It sounds awesome.

Firstyear:
Anyway, with those changes I'm really happy to accept this. Thanks so much for the PR!

adrienperonnet:
Awesome. Thanks for your feedbacks @Firstyear 👍

src/lib.rs
adrienperonnet: Yes this is very neat. I think it's working out of the box if you have a mac with TouchID hardware and an up to date browser. You can see a full demo here : https://drive.google.com/file/d/1IyPJ8X4MBOLLffut4yXXz1409s6ukNGC/view

Firstyear: That's extremely cool, I do have a touchid mac, but I'm using FF not Chrome so I didn't know this was possible!

PR #20, opened 2020-01-09
fix tokenBinding client data type

agrinman:
TokenBinding is struct (not a string): https://www.w3.org/TR/webauthn/#collectedclientdata-json-serialized-client-data

- [ ] cargo fmt has been run

### DIFF
--- a/src/proto.rs
+++ b/src/proto.rs
@@ -245,7 +245,7 @@ pub(crate) struct CollectedClientData {
     pub(crate) type_: String,
     pub(crate) challenge: Vec<u8>,
     pub(crate) origin: String,
-    pub(crate) tokenBinding: Option<String>,
+    pub(crate) tokenBinding: Option<TokenBinding>,
 }
 
 #[derive(Debug, Deserialize)]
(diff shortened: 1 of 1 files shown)


Firstyear:
Hey there, thanks for picking up on this mistake. 

Is it also possible to have an example in the tests for parsing this? Is there a specific type of token or device that uses this field?

agrinman:
I believe yubikeys on Edge are the only combo that support it. Still though, the structure should parse correctly.

Firstyear:
Hmmm I was more asking if you could submit a unit test containing the flags/vectors associated so that we can guarantee that we don't break the use case. I don't have access to edge to create these is all :) 

Can you generate the test vectors and add them to a test case? 

Thanks again for your submission and patience :)

agrinman:
I actually also don't have access to one. There also aren't any existing tests for token binding. I think a more correct PR is just to remove the tokenBinding field altogether (until someone has an actual use case for it, and can PR with tests/motivation). 

The reason I found this bug because the [Krypton extension also had a legacy/incorrect tokenBinding structure/field](https://github.com/kryptco/kr-u2f/pull/30) and the webauthn-rs was failing on parsing the input.

Firstyear:
Okay, I'm going to merge this, and make it my problem to add tests in the future. Thanks for the contribution. If you have some time, please also look at issue https://github.com/Firstyear/webauthn-rs/issues/19 about ergonomics of the library, as I'd appreciate any feedback you have!

Issue #19, opened 2019-10-21
Review public API


Firstyear:
An open question is the ergonomics of the public API and how we expose datastorage traits and other methods of application integration. As this library hasn't been used in "production" yet, it's still an open question if the decisions I made were really correct or ergonomic to use for deployments.

If you have comments or feedback on the current API design, I'd love to know!

Firstyear:
cc @adrienperonnet since you have been looking at this library recently.

agrinman:
@Firstyear Thanks for creating this project! Great work and overall it was very easy to use. 

Re: ergonomics the one feedback I have is that it might be a bit too opinionated with the WebAuthnConfig trait. I had to fight the abstraction a bit by creating a struct that wraps a mutable reference of the main struct that implements the trait. For example, I needed to pull out some data after registration/auth was performed (like which credential it was that succeeded). In some cases you might store challenges in a session cookie or in other cases a database.

What I imagine could be a less opinionated API is something like ~5 functions:
-`generate_challenge`
-`register_request`
-`complete_register`
-`authenticate_request`
-`complete_authenticate`

(Naming could be more symmetric to the actual webauthn spec, but you get the idea.)

Firstyear:
I can see where you are coming from. You want the api to simply do the cryptographic operations given some json, you get "an output".

I think the issue with this is that on each operation then, you would need origin, relying party name, party id, valid algos on each request.

So you would probably end up with something where you have to give a "webauthnconfig" on each function call instead, where as the current model tries to have the "policy" defined once and then reusable in a thread safe/wrapped manner. 

Don't think I'm saying no to the suggestion - but I'd like to get the idea a bit better defined before I commit to it :) 


Some example cases is that on generate_challenge, we need to know:

* the credentials associated to a username
* if verified tokens are requested or not
* the timeout window
* the current relying party id
* a way to store the challenge

So I'd imagine for your example the function might change to say;

```
fn generate_challenge_authenticate(&config: WebauthnConfig, username: UserId, credentials: &[Credential]) -> WebauthnChallenge
```

This would make it such that:

* The config between calls to the function could change (good or bad, depends on your view). We still have a config too that has to do quite a bit of heavy lifting ... 
* you have to externally provide the usercredentials
* you have to persist the challenge associated to a session/transaction.

[...]

PR #21, opened 2020-02-09
20200207 update webauthn api

Firstyear:
Implements #19 improve api ergonomics .

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -4,7 +4,7 @@
 //! to allow strong, passwordless, cryptographic authentication to be performed. Webauthn
 //! is able to operate with many authenticator types, such as U2F.
 //!
-//! This library aims to provide a secure black-box Webauthn implementation that you can
+//! This library aims to provide a secure Webauthn implementation that you can
 //! plug into your application, so that you can provide Webauthn to your users.
 //!
 //! For examples, see our examples folder.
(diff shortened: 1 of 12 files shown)


src/lib.rs
```diff
@@ -101,7 +113,7 @@ impl<T> Webauthn<T> {
         Webauthn {
             // We use stdrng because unlike thread_rng, it's a csprng, which given
```
agrinman: Outdated comment?
`thread_rng` is cryptographically secure and it is what's being used.

src/lib.rs
```diff
@@ -509,27 +536,90 @@ impl<T> Webauthn<T> {
...
+    ///
+    /// You must persist the AuthenticationState that is returned. You should associate this by
+    /// UserId. The AuthenticationState is required for the authenticate_credential function to
+    /// operate correctly.
+    ///
+    /// At this time we deviate from the standard and base64 some fields, but we are
+    /// investigating how to avoid this (https://github.com/Firstyear/webauthn-rs/issues/5)
+    pub fn generate_challenge_authenticate(
+        &mut self,
+        _username: &UserId,
```
agrinman: It looks like this field is unnecessary?

src/lib.rs
Firstyear: Fixed

src/lib.rs
Firstyear: And removed :)

Issue #14, opened 2019-05-01
base64 in demo site only works if username > len 2


Firstyear:

Firstyear:
Resolved with wasm + recent updates.

Issue #5, opened 2019-04-20
Investigate if it is possible for JSON to avoid base64 in js/json


Firstyear:
We currently deviate from the standard by base64 encoding some JSON members in challenges and responses. This may confuse some consumers, but this difficulty comes in from ArrayBuffers being a `'a [u8]` in a struct, which causes lifetime issues with actix. I have also noted odd behaviours where the output of navigator.credentials methods can't be sent to clients either, also causing us to do some odd js tricks.

Basically, we need to improve the js parts, and work out better ways to handle this aspect of the system.

aseigler:
For conformance particularly, assume everyone wants base64url encoding without padding.

Firstyear:
The spec was unclear (or I missed it?) about the base64 as url-safe or not. At the moment I attempt both url and standard, and fail only if both fail. I've opened a seperate issue to handle this issue.

This issue is more about the fact that there are a number of ArrayBuffer types that I just can't seem to handle nicely with Serde, so as a result my client js and server are base64'ing data that perhaps should not be. For example, the rawId in authenticate I am base64 encoding, making it equivalent to Id field.

So to be properly compliant with "any javascript client" parts, I should not require base64 on these arrayBuffer types incorrectly. :)

Firstyear:
Closing because it's not possible to create json types that can become a Uint8Array on a browser. You *must* base64 these datafields. It surprises me that the webauthn standard does not just say "base64 these" and then leave it to the navigator.credentials api to handle it, but alas, instead all javascript implementations now have the extreme joy of working out how to get from base64 to a uint8Array and tampering with the internals of any server issued request.

PR #25, opened 2020-05-17
W3C Conformance Fixes and Other Minor Improvements

agrinman:
A few minor improvements:
- allow library user customization on register via arguments 
- use snake case and serde rename
- use a custom type to represent base64 url safe, no pad encoded bytes (implement ser/de traits on it)
- expose certain properties to library users (i.e. pub(crate) -> pub).
- WebAuthnError conforms to Error trait now via `this error`.

- [ ] cargo fmt has been run
- [x ] cargo test has been run and passes
- [x ] documentation has been updated with relevant examples (if relevant)

### DIFF
--- a/examples/actix/main.rs
+++ b/examples/actix/main.rs
@@ -229,7 +229,7 @@ fn main() {
     let domain = opt.rp_id.clone();
 
     // Generate TLS certs as needed.
-    let ssl_params = generate_dyn_ssl_params(domain.as_str());
+    // let ssl_params = generate_dyn_ssl_params(domain.as_str());
 
     let wan_c = WebauthnEphemeralConfig::new(
         opt.rp_name.as_str(),
@@ -277,7 +277,9 @@ fn main() {
             )
     });
     server
-        .bind_openssl(opt.bind.as_str(), ssl_params)
+        // .bind_openssl(opt.bind.as_str(), ssl_params)
+        // .unwrap()
+        .bind("localhost:8080")
         .unwrap()
         .run();
 
--- a/examples/actix/main.rs
+++ b/examples/actix/main.rs
@@ -58,15 +58,17 @@ struct CmdOptions {
     #[structopt(
         short = "n",
         long = "name",
-        default_value = "https://localhost:8443/auth"
+        default_value = "http://localhost:8080/auth"
     )]
     rp_name: String,
-    #[structopt(short = "o", long = "origin", default_value = "https://localhost:8443")]
+    #[structopt(short = "o", long = "origin", default_value = "http://localhost:8080")]
     rp_origin: String,
     #[structopt(short = "i", long = "id", default_value = "localhost")]
     rp_id: String,
-    #[structopt(short = "b", long = "bind", default_value = "127.0.0.1:8443")]
+    #[structopt(short = "b", long = "bind", default_value = "localhost:8080")]
     bind: String,
+    #[structopt(short = "s", long = "tls")]
+    enable_tls: bool,
 }
 
 // fn index_view(session: Session) -> HttpResponse {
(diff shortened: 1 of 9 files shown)


examples/actix/main.rs
```diff
@@ -229,7 +229,7 @@ fn main() {
     let domain = opt.rp_id.clone();
 
     // Generate TLS certs as needed.
-    let ssl_params = generate_dyn_ssl_params(domain.as_str());
+    // let ssl_params = generate_dyn_ssl_params(domain.as_str());
```
Firstyear, line 232: If you remove this then the localhost test server won't work because the browser won't send over non-https IIRC.

src/proto.rs
```diff
@@ -8,9 +8,12 @@ use std::convert::TryFrom;
 
 use crate::crypto;
 use crate::error::*;
+use crate::base64_data::Base64UrlSafeData;
+
+use serde::{Serialize, Deserialize};
 
 /// Representation of a UserId. This is currently a type alias to "String".
```
Firstyear, line 15: Minor, but probably worth updating the doc string

src/proto.rs
```diff
@@ -147,72 +134,95 @@ pub(crate) struct AllowCredentials {
...
+pub struct AuthenticatorSelectionCriteria {
+    authenticator_attachment: Option<String>,
+    require_resident_key: Option<bool>,
+    user_verification: Option<String>
+}
+
+
+/// https://www.w3.org/TR/webauthn/#enumdef-attestationconveyancepreference
+#[derive(Debug, Serialize)]
+#[serde(rename_all="lowercase")]
```
Firstyear, line 164: That's magic, very cool!

Firstyear:
Besides some really minor nitpicks, this change is fantastic and fixes a number of places that I knew were problems but didn't yet understand how to solve. Thanks very much!

Firstyear:
I think this will resolve:

fixes #14 
fixes #10 
fixes #5 

Is it worth submitting your base64 url safe serde decoder to the base64 crate too so we can just rely on it from there? Not that I mind it being here.

agrinman:
Thanks for the review. I'll address the suggestions shortly in a new commit.

Re: base64, there's this https://docs.rs/base64-serde/0.5.1/base64_serde/ that we could use instead. The one reason I see for actually keeping our own type, is that when we decode we should make it accept UrlSafe, UrlsSafeNoPad, and Standard (because why not?) but when we encode we use UrlSafeNoPad as that's only what the webauthn api supports. It's up to the javascript client to turn "some" of those fields into byte arrays. A more correct solution could use different base64data types to match what fields the webauthn api expects to be strings versus byte arrays...but may be over kill for now. I opted just for consistency for now.

examples/actix/main.rs
agrinman: From:

https://developer.mozilla.org/en-US/docs/Web/API/Web_Authentication_API

> Please note: both create() and get() require a Secure Context (e.g. - the server is connected by https or is the localhost), and will not be available for use if the browser is not operating in a secure context.

I verified it works for me on localhost:<port>. Might you try again and see if it fails for you? Maybe one browser allows it and another doesn't?

Firstyear: I remember what it was, it was when I was testing from my ipad to my development machine, that's why it needed the https generator. What's the main reason to disable it?

Firstyear:
That makes sense, I'm happy to have our own type, and consistency matters. So once you comment about the https thing, I'm happy to merge this. Thanks! Did you want me to do a release after this as I assume you want to be able to use these changes?

examples/actix/main.rs
agrinman: I got weird issues on safari when the certificate changed, it wouldn't let me skip pass the warning. in general, it's a bit harder to test with https locally (various browsers do different things here). We can meet in the middle and just make it an option at run-time, wdyt?

Firstyear: Yep, and I'm happy for it to default to "off", so that only when I or others need some esoteric test they can enable it. Thanks!

Firstyear:
Thank you very much! Is there anything else you wanted to add, or do you want me to do a release so you can use the changes?

PR #35, opened 2020-11-27
feat(tide-example): working tide example

ericmarkmartin:
Implements # .

- [x] cargo fmt has been run


Cargo.toml
```diff
@@ -32,11 +32,19 @@ thiserror = "1.0"
...
 actix-web = { version = "2.0", features = ["openssl"] }
 actix-session = "0.3"
 actix-files = "0.2"
 
+# Example, use the version numbers you need
+tide = "0.15.0"
+tide-rustls = "0.1.3"
+askama_tide = "0.12.0"
+async-std = { version = "1.6.0", features = ["attributes"] }
+xactor = "0.7.8"
```
Firstyear, line 45: I think it would be good here to remove xactor and actix completely if we are going to jump to tide :)

examples/tide/actors.rs
```diff
@@ -0,0 +1,170 @@
+use webauthn_rs::ephemeral::WebauthnEphemeralConfig;
+use webauthn_rs::error::WebauthnError;
+use webauthn_rs::proto::{
+    CreationChallengeResponse, Credential, CredentialID, PublicKeyCredential,
+    RegisterPublicKeyCredential, RequestChallengeResponse, UserId, UserVerificationPolicy,
+};
+use webauthn_rs::{AuthenticationState, RegistrationState, Webauthn};
+use xactor::*;
```
Firstyear, line 8: Rather than xactor I think we can use async-std with task::spawn

Cargo.toml
```diff
@@ -32,11 +32,19 @@ thiserror = "1.0"
...
 # This only supports up to 0.7. See examples/actix/main.rs for more.
-actix = "0.9"
+actix = "0.10"
 actix-web = { version = "2.0", features = ["openssl"] }
 actix-session = "0.3"
 actix-files = "0.2"
 
+# Example, use the version numbers you need
+tide = "0.15.0"
+tide-rustls = "0.1.3"
```
Firstyear, line 42: In versions I think it's best not to hardcode the patch release so `tide-rustls = "0.1.3"` should be `tide-rustls = "0.1"` so that cargo update works when new fixes come out. Similar for the other added deps.

examples/tide/README.md
```diff
@@ -0,0 +1,30 @@
+Actix with Webauthn
```
Firstyear, line 1: This text needs to be updated

src/lib.rs
```diff
@@ -118,10 +118,10 @@ impl<T> Webauthn<T> {
         let rp_id_hash = compute_sha256(config.get_relying_party_id().as_bytes());
         Webauthn {
             // Use a per-thread csprng
-            rng: rand::thread_rng(),
-            config: config,
-            pkcp: pkcp,
-            rp_id_hash: rp_id_hash,
+            rng: rand::rngs::StdRng::seed_from_u64(42),
```
Firstyear, line 121: This is sadly, a security issue. By setting a static seed it means challenges are predictable. I have a better fix for this in my own branch that I'll push shortly, so you can probably consider at least reverting the changes here in lib.rs.

Firstyear (changes requested):
Thanks for adding this, some comments have been added. Really appreciate your time :)

Firstyear:
Please note, I'm pushing the breaking changes to master now for 0.3.0 which means youll need to rebase, but it should fix the rng issue you had :)

Cargo.toml
```diff
@@ -31,11 +31,11 @@ thiserror = "1.0"
...
 askama = "0.10"
 
-# This only supports up to 0.7. See examples/actix/main.rs for more.
-actix = "0.9"
-actix-web = { version = "2.0", features = ["openssl"] }
-actix-session = "0.3"
-actix-files = "0.2"
+# Example, use the version numbers you need
+tide = "0.15"
+# tide-rustls = "0.1"
```
ericmarkmartin, line 36: Not sure what you want to do about this—I think we can either use `rustls` in the example or just not offer a TLS option.

examples/tide/main.rs
```diff
@@ -0,0 +1,186 @@
...
+            tide::Response::new(tide::StatusCode::InternalServerError)
+        }
+    };
+    Ok(res)
+}
+
+async fn register(mut request: tide::Request<AppState>) -> tide::Result {
+    let username = request.param("username")?.parse()?;
+    let reg = request.body_json::<RegisterPublicKeyCredential>().await?;
+    let actor_res = request.state().register(&username, &reg).await;
```
ericmarkmartin, line 91: We could kick these futures off with `async_std::task::spawn` but everything in tide ultimately gets run in the async-std runtime anyway so I'm not sure it makes a difference.

examples/tide/main.rs
```diff
@@ -0,0 +1,186 @@
...
+    app.at("/auth/challenge/register/:username")
+        .post(challenge_register);
+    app.at("/auth/challenge/login/:username")
+        .post(challenge_login);
+    app.at("/auth/register/:username").post(register);
+    app.at("/auth/login/:username").post(login);
+    app.at("/auth/static/").serve_dir("static")?;
+
+    // TODO: tide_rustls uses version 0.14.0 (we're on 0.15.0) and uses rustls over openssl
+    // if opt.enable_tls {
```
ericmarkmartin, line 173: See above comment about tls

Cargo.toml
Firstyear: We can use tide-rustls, we need a TLS option for some development / testing use cases. :)

examples/tide/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+WORKDIR /src
+
+ADD ./ /src/
+
+RUN zypper in -y gcc libopenssl-devel openssl wget && \
+    wget https://static.rust-lang.org/rustup/dist/x86_64-unknown-linux-gnu/rustup-init && \
+    chmod +x rustup-init && \
+    ./rustup-init -v -y && \
+    source /root/.profile && \
+    cargo build --example actix --release && \
```
Firstyear, line 16: Shouldn't this be example tide?

examples/tide/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+    zypper rm -y gcc libopenssl-devel && \
+    rm -r rustup-init /root/.rustup
+
```
Firstyear, line 19: There is a way to build this in an intermediate container and then copy the binary over. Have a look at https://github.com/kanidm/kanidm/blob/master/kanidmd/Dockerfile#L34

examples/tide/actors.rs
```diff
@@ -0,0 +1,163 @@
...
+use webauthn_rs::error::WebauthnError;
+use webauthn_rs::proto::{
+    CreationChallengeResponse, Credential, CredentialID, PublicKeyCredential,
+    RegisterPublicKeyCredential, RequestChallengeResponse, UserId, UserVerificationPolicy,
+};
+use webauthn_rs::{AuthenticationState, RegistrationState, Webauthn};
+
+use lru::LruCache;
+use std::collections::BTreeMap;
+use std::sync::Mutex;
```
Firstyear, line 11: This needs to be an async mutex else we could deadlock,

examples/tide/main.rs
Firstyear: I don't think it matters to much in this case, I think maybe task::spawn is over-kill. But as mentioned we need async mutex

examples/tide/main.rs
Firstyear: Just use tide 0.14 then :)

Firstyear:
looking much better, it's really getting there! Would be good to rm the actix folder too if you want since this will replace it :) 

Also feel free to add yourself to the contributors.md file too since you are doing this work

Cargo.toml
ericmarkmartin: Do you want me to generate a test cert?

Firstyear: That's what the related openssl code does, it generates a self signed localhost cert on the fly

examples/tide/Dockerfile
```diff
@@ -13,10 +13,12 @@ RUN zypper in -y gcc libopenssl-devel openssl wget && \
     chmod +x rustup-init && \
     ./rustup-init -v -y && \
     source /root/.profile && \
-    cargo build --example actix --release && \
+    cargo build --example tide --release && \
     zypper rm -y gcc libopenssl-devel && \
     rm -r rustup-init /root/.rustup
 
-CMD ['/src/target/release/examples/actix']
+COPY --from=builder /usr/src/target/release/examples/tide /sbin/
```
Firstyear, line 20: I think this is missing a few steps. You need a second FROM tag for the second part. Note L24 and L2 in https://github.com/kanidm/kanidm/blob/master/kanidmd/Dockerfile :) 

did you try to build it too?

Cargo.toml
ericmarkmartin: Right, but I think `tide-rustls` will only accept the cert and key via a file path. Would you prefer I make a static cert and key or use the on the fly strategy and serialize to a temp directory?

examples/tide/Dockerfile
ericmarkmartin: Hopefully I pattern matched this correctly—I'm not experienced with Docker.

examples/tide/main.rs
ericmarkmartin: Fixed

examples/tide/main.rs
ericmarkmartin: Oh this is fun. Tide 0.14 doesn't have `with_cookie_domain` on the `SessionMiddleware` object.

Cargo.toml
Firstyear: Probably a much bigger task than you bargained for - You can use `.config()` with a rustls `ServerConfig`, and you can feed that your key/cert with https://docs.rs/rustls/0.19.0/rustls/internal/pemfile/fn.rsa_private_keys.html. BufRead is implemented by &[u8], so that means you should be able to take the generated cert as a string, do as_bytes() and then feed that to this function to provide the private key and certs in memory. 

You did way you wanted to be challenged, right ;)

examples/tide/main.rs
Firstyear: For now, we can just remove it. it's not the most critical thing in this application.

examples/tide/Dockerfile
Firstyear: It's *prettttyyyyyyyyy* close to what it needs. At this point I'm happy for you to leave it as is, and I will fix it later :)

Cargo.toml
ericmarkmartin: This doesn't seem too bad—mostly like just reimplementing tide-rustls. Before I begin such an undertaking though, I want to raise two possible alternatives

1. PR tide-rustls to accept certs and keys in memory
2. Use something like [tempfile](https://docs.rs/tempfile/3.1.0/tempfile/) to create temporary certs and keys.

examples/tide/main.rs
ericmarkmartin: Ah it looks like tide-rustls got updated to the correct version since I PRd.

examples/tide/actors.rs
```diff
@@ -0,0 +1,155 @@
+use webauthn_rs::ephemeral::WebauthnEphemeralConfig;
```
ericmarkmartin, line 1: I'm curious if you have a better name for this than Actor, now that it's not an actor at all.

examples/tide/Dockerfile
ericmarkmartin: I tried running it from the tide directory with `docker build .` and it seems to only include the example directory, rather than the whole repo (which is what the `cargo build` command seems to suppose). Because the actix dockerfile had an identical add command, I take it I'm building the container incorrectly?

Cargo.toml
ericmarkmartin: I think using OpenSSL's [private_key_to_der](https://docs.rs/openssl/0.10.30/openssl/pkey/struct.PKey.html#method.private_key_to_der) might work as well.

ericmarkmartin: Done—not so bad

examples/tide/Dockerfile
ericmarkmartin: I actually couldn't even get `docker build .` to succeed with the old actix example seemingly because only the example directory, rather than the whole repo, was being added to the container.

ericmarkmartin:
> looking much better, it's really getting there! Would be good to rm the actix folder too if you want since this will replace it :)
> 
> Also feel free to add yourself to the contributors.md file too since you are doing this work

Also, thanks for this comment, but I don't see a contributors.md file in the repo.

Firstyear:
Ahh feel free to add it then, with the same format as https://raw.githubusercontent.com/kanidm/concread/master/CONTRIBUTORS.md

examples/tide/Dockerfile
Firstyear: I think the file should look like this:

```
ARG BASE_IMAGE=opensuse/leap:latest
FROM ${BASE_IMAGE} AS builder
MAINTAINER wbrown@suse.de

RUN mkdir /src
WORKDIR /src

ADD ./ /src/

RUN zypper in -y gcc libopenssl-devel openssl wget && \
    wget https://static.rust-lang.org/rustup/dist/x86_64-unknown-linux-gnu/rustup-init && \
    chmod +x rustup-init && \
    ./rustup-init -v -y && \
    source /root/.profile && \
    cargo build --example tide

FROM ${BASE_IMAGE}
LABEL mantainer william@blackhats.net.au

RUN zypper ref && \
    zypper install -y \
        timezone && \
    zypper clean -a

COPY --from=builder /src/target/debug/examples/tide /sbin/

EXPOSE 8080
CMD ['/sbin/tide']

```

examples/tide/actors.rs
Firstyear: The concept is still pretty similar so I'd just leave it for now.

Firstyear:
Up to you if you want to apply the changes suggested re the dockerfile but at this point I'm happy to merge it :)

examples/tide/Dockerfile
ericmarkmartin: I tried something like this and it almost works—I also had to copy over the static directory though. Ultimately I didn't get anything on localhost:8000 when I tried to run the container but I'm trying to see if that can be fixed simply.

Firstyear: I think maybe add

```
COPY --from=builder /src/static /static
WORKDIR /
```

That may fix it.

Issue #23, opened 2020-02-09
Update actix+actix web


Firstyear:
Update these to the latest versions, and improve the json debugging capability.

PR #37, opened 2020-11-30
Update actix_web example to version 3

PascalKoe:
Implements #23 

- [x] cargo fmt has been run
- [X] cargo test has been run and passes
- [X] documentation has been updated with relevant examples (if relevant)

Don't know if you are still interested in an actix_web example with @ericmarkmartin putting work into a tide example.
Feel free to rip this code apart as I would be surprised if my more or less first rust code is any good :)

Because I'm a bit of a noob and had a hard time understanding what was going on in the original actix_web example, I went more on the verbose side of variables/functions/comments.


Firstyear:
Hey there @PascalKoe. The reason why @ericmarkmartin is doing the tide example is because I want to remove actix from all my associated projects, so as much as I really appreciate your work on this, I can't accept it because we are just about to remote this. Of course, I'd love to help out if there are other contributions you wish to make, thank you!

PR #39, opened 2020-11-30
add redirect to /auth from /

ericmarkmartin:
Implements # .

- [x] cargo fmt has been run

### DIFF
--- a/examples/tide/main.rs
+++ b/examples/tide/main.rs
@@ -159,6 +159,11 @@ async fn main() -> tide::Result<()> {
             .with_cookie_name("webauthnrs");
     app.with(sessions);
     app.with(tide::log::LogMiddleware::new());
+    {
+        let prefix_copy = prefix.clone();
+        app.at("/")
+            .get(move |_| async_std::future::ready(Ok(tide::Redirect::new(prefix_copy.clone()))));
+    }
     app.at(&prefix).get(index_view);
     app.at(&format!("{}/challenge/register/:username", prefix))
         .post(challenge_register);


Firstyear:
I think you need to rebase this with your other prefix change, but it looks good!

examples/tide/main.rs
```diff
@@ -159,6 +159,11 @@ async fn main() -> tide::Result<()> {
             .with_cookie_name("webauthnrs");
     app.with(sessions);
     app.with(tide::log::LogMiddleware::new());
+    {
+        let prefix = prefix.clone();
+        app.at("/")
+            .get(move |_| async_std::future::ready(Ok(tide::Redirect::new(prefix.clone()))));
```
Firstyear, line 165: You don't need this extra clone here since you have the prefix clone above.

ericmarkmartin: I believe without the extra clone we move ownership out of prefix which causes issues when we use it below.

Firstyear: To make this clearer:

```
let prefix_copy = prefix.clone()
app.at().get(move |_| ... new(prefix_copy) 
```

This is clearer becuase it doesn't rely on shadowing the variable name, but also because the .get closure is move, then prefix copy will be moved there, meaning the second clone isn't needed, and the original prefix value is not affected.

ericmarkmartin: Oh sorry, I see you were talking about the clone within the closure. I believe that clone is necessary so the closure satisfies `Fn`, otherwise it only satisfies `FnOnce`.

Firstyear: Do you mind pasting the compiler error?

ericmarkmartin:
```
error[E0525]: expected a closure that implements the `Fn` trait, but this closure only implements `FnOnce`
   --> examples/tide/main.rs:164:14
    |
164 |         .get(move |_| async_std::future::ready(Ok(tide::Redirect::new(prefix_copy))));
    |          --- ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^-----------^^^
    |          |   |                                                        |
    |          |   |                                                        closure is `FnOnce` because it moves the variable `prefix_copy` out of its environment
    |          |   this closure implements `FnOnce`, not `Fn`
    |          the requirement to implement `Fn` derives from here
```

Firstyear: Well today I learnt. I'd still prefer you avoid the shadow naming, so would be good to rename let prefix to let prefix_copy or similar.

ericmarkmartin: pushed

Firstyear:
Thanks again for adding this!

Issue #15, opened 2019-05-01
zero length usernames should be rejected


Firstyear:

ericmarkmartin:
Is there an existing error for this in `src/error.rs` or should a new one be made?

Firstyear:
I'd add a InvalidUsername error for this :)

ericmarkmartin:
Should this issue be closed now?

PR #40, opened 2020-12-02
reject empty userids in registration

ericmarkmartin:
Implements #15.

- [x] cargo fmt has been run

### DIFF
--- a/examples/tide/main.rs
+++ b/examples/tide/main.rs
@@ -56,8 +56,13 @@ async fn index_view(mut request: tide::Request<AppState>) -> tide::Result {
     Ok(askama_tide::into_response(&IndexTemplate {}, "html"))
 }
 
+fn username_from_request(request: &tide::Request<AppState>) -> String {
+    request.param("username").unwrap_or("").to_string()
+}
+
 async fn challenge_register(request: tide::Request<AppState>) -> tide::Result {
-    let username = request.param("username")?.parse()?;
+    // let username = request.param("username")?.parse()?;
+    let username = username_from_request(&request);
     let actor_res = request.state().challenge_register(username).await;
     let res = match actor_res {
         Ok(chal) => tide::Response::builder(tide::StatusCode::Ok)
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -1565,4 +1569,33 @@ mod tests {
         println!("{:?}", result);
         assert!(result.is_ok());
     }
+
+    use super::*;
```
Firstyear, line 1573: Better to avoid blanket imports

examples/tide/main.rs
```diff
@@ -56,8 +56,13 @@ async fn index_view(mut request: tide::Request<AppState>) -> tide::Result {
     Ok(askama_tide::into_response(&IndexTemplate {}, "html"))
 }
 
+fn username_from_request(request: &tide::Request<AppState>) -> String {
+    request.param("username").unwrap_or("").to_string()
+}
+
 async fn challenge_register(request: tide::Request<AppState>) -> tide::Result {
-    let username = request.param("username")?.parse()?;
+    // let username = request.param("username")?.parse()?;
```
Firstyear, line 64: No need to leave this commented

examples/tide/main.rs
```diff
@@ -56,8 +56,13 @@ async fn index_view(mut request: tide::Request<AppState>) -> tide::Result {
     Ok(askama_tide::into_response(&IndexTemplate {}, "html"))
 }
 
+fn username_from_request(request: &tide::Request<AppState>) -> String {
```
Firstyear, line 59: Is there a reason this has to be a function? It's only used in one place ...

ericmarkmartin: Sorry—I just forgot to clean this up. This was me playing around to see if I could get the error in the example, but I wasn't able to because the tide router is smart enough to not route `/register/` to the route set to `/register/:username`.

src/lib.rs
```diff
@@ -1565,4 +1569,34 @@ mod tests {
         println!("{:?}", result);
         assert!(result.is_ok());
     }
+
+    use super::{CreationChallengeResponse, RegistrationState, WebauthnError};
```
Firstyear, line 1573: Shoud be on line 924, and use crate not super.

Firstyear (changes requested):
Sorry, found one more little thing, otherwise it looks great!

PR #43, opened 2020-12-10
fix(policy_verify_trust): update documentation to match default impl

### LINKED ISSUE #42: Default implementation of `policy_verify_trust` doesn't match documentation
The doc comment above the default implementation of `policy_verify_trust` says

```
/// The default implementation of this method rejects None and Uncertain attestation, and
/// will "blindly trust" self attestation and the other types as valid.
/// If you have strict security requirements we strongly recommend you implement this function,
/// and we may in the future provide a stronger default relying party policy.
```

The implementation, however, appears to accept the `None` attestation type
```
fn policy_verify_trust(&self, at: AttestationType) -> Result<Credential, ()> {
    log::debug!("policy_verify_trust -> {:?}", at);
    match at {
        AttestationType::Basic(credential, _attest_cert) => Ok(credential),
        AttestationType::Self_(credential) => Ok(credential),
        AttestationType::AttCa(credential, _attest_cert, _ca_chain) => Ok(credential),
        AttestationType::None(credential) => Ok(credential),
        _ => {
            // We don't know how to assert trust in this yet, or we just
            // don't trust it at all (Uncertain, None).
            Err(())
        }
    }
}
```

ericmarkmartin:
Implements #42.

- [x] cargo fmt has been run

Fixes #42.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -904,22 +904,20 @@ pub trait WebauthnConfig {
     /// If the attestation statement attStmt successfully verified but is not trustworthy per step
     /// 16 above, the Relying Party SHOULD fail the registration ceremony.
     ///
-    /// The default implementation of this method rejects None and Uncertain attestation, and
+    /// The default implementation of this method rejects ECDAA and Uncertain attestation, and
     /// will "blindly trust" self attestation and the other types as valid.
     /// If you have strict security requirements we strongly recommend you implement this function,
     /// and we may in the future provide a stronger default relying party policy.
     fn policy_verify_trust(&self, at: AttestationType) -> Result<Credential, ()> {
         log::debug!("policy_verify_trust -> {:?}", at);
         match at {
-            AttestationType::Basic(credential, _attest_cert) => Ok(credential),
-            AttestationType::Self_(credential) => Ok(credential),
-            AttestationType::AttCa(credential, _attest_cert, _ca_chain) => Ok(credential),
-            AttestationType::None(credential) => Ok(credential),
-            _ => {
-                // We don't know how to assert trust in this yet, or we just
-                // don't trust it at all (Uncertain, None).
-                Err(())
-            }
+            AttestationType::Basic(credential, _)
+            | AttestationType::Self_(credential)
+            | AttestationType::AttCa(credential, _, _)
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -904,22 +904,20 @@ pub trait WebauthnConfig {
     /// If the attestation statement attStmt successfully verified but is not trustworthy per step
     /// 16 above, the Relying Party SHOULD fail the registration ceremony.
     ///
-    /// The default implementation of this method rejects None and Uncertain attestation, and
+    /// The default implementation of this method rejects ECDAA and Uncertain attestation, and
```
Firstyear, line 907: I'd make it just "rejects Uncertain" because when we add support for ECDAA we'll "not" blindly reject it.

src/lib.rs
```diff
@@ -904,22 +904,20 @@ pub trait WebauthnConfig {
...
-            AttestationType::AttCa(credential, _attest_cert, _ca_chain) => Ok(credential),
-            AttestationType::None(credential) => Ok(credential),
-            _ => {
-                // We don't know how to assert trust in this yet, or we just
-                // don't trust it at all (Uncertain, None).
-                Err(())
-            }
+            AttestationType::Basic(credential, _)
+            | AttestationType::Self_(credential)
+            | AttestationType::AttCa(credential, _, _)
```
Firstyear, line 916: I'd rather leave this "as they were" as a copy-paste example with the correct variable names to what each element of the enum is.

Firstyear (changes requested):

src/lib.rs
```diff
@@ -904,20 +904,21 @@ pub trait WebauthnConfig {
     /// If the attestation statement attStmt successfully verified but is not trustworthy per step
     /// 16 above, the Relying Party SHOULD fail the registration ceremony.
     ///
-    /// The default implementation of this method rejects ECDAA and Uncertain attestation, and
+    /// The default implementation of this method rejects None and Uncertain attestation, and
```
Firstyear: rejects Uncertain attestation :)

Firstyear (changes requested):

Firstyear:
Thank you!

PR #47, opened 2021-02-15
Several new features and edge case fixes

agrinman:
- more configuration values to support edge cases
	- allow skipping counter verification
	- allow unknown attestation formats
- return authenticator data (not only counter) and expose extension outputs
- support user-less mode (mutate authentication state to add credentials)
- other misc fixes
- add GitHub Actions (free) CI

- [x] cargo fmt has been run


agrinman:
Future follow up: I think a good improvement to this code base is to separate out some of the modules into separate files. Some of the modules contain > 1k lines which makes reasoning about it difficult.

src/core.rs
```diff
@@ -769,33 +787,29 @@ impl<T> Webauthn<T> {
...
+        // greater than the signature counter value stored in conjunction with credential’s id attribute.
+        //       Update the stored signature counter value, associated with credential’s id attribute,
+        //       to be the value of authData.signCount.
+        // less than or equal to the signature counter value stored in conjunction with credential’s id attribute.
+        //      This is a signal that the authenticator may be cloned, i.e. at least two copies
+        //      of the credential private key may exist and are being used in parallel. Relying
+        //      Parties should incorporate this information into their risk scoring. Whether the
+        //      Relying Party updates the stored signature counter value in this case, or not,
+        //      or fails the authentication ceremony or not, is Relying Party-specific.
+        let counter_shows_compromise = auth_data.counter <= cred.counter;
```
Firstyear, line 806: nit pick, we could do this inside of an 'if check_counter', but that's a super minor point.

Firstyear (commented):
Generally this looks awesome! One minor nitpick in comments. It's a weird requirement, but can you check this works on rust 1.45.0 as well? 

I also agree, a bit refactor/cleanup is needed too. Maybe open an issue so we don't forget about it :) 

Thanks very much!

Firstyear:
Awesome! Thank you so much! Sorry about the 1.45 thing, it's just that's the version we are stuck on at SUSE for some things and Webauthn-rs gets pulled to kanidm that has to support that minimum version. *sigh*.

Anyway, great work, thank you!

agrinman:
Thanks! 
Re: 1.45, I've added a build job in actions that builds against it: https://github.com/agrinman/webauthn-rs/runs/1900700137?check_suite_focus=true

PR #49, opened 2021-02-18
User verification logic

agrinman:
Re: the discussion here: https://github.com/kanidm/webauthn-rs/commit/9e3fd65de11594955790cc1ebb60e567398f0cfd

What do you think of this approach for dealing with the UV problem?

### DIFF
--- a/src/core.rs
+++ b/src/core.rs
@@ -53,7 +53,6 @@ impl AuthenticationState {
     }
 }
 
-
 /// This is the core of the Webauthn operations. It provides 4 interfaces that you will likely
 /// use the most:
 /// * generate_challenge_response
@@ -413,6 +412,7 @@ impl<T> Webauthn<T> {
             AttestationFormat::FIDOU2F => verify_fidou2f_attestation(
                 acd,
                 data.attestation_object.auth_data.counter,
+                policy,
                 data.attestation_object.auth_data.user_verified,
                 &data.attestation_object.att_stmt,
                 &client_data_json_hash,
@@ -421,6 +421,7 @@ impl<T> Webauthn<T> {
             AttestationFormat::Packed => verify_packed_attestation(
                 acd,
                 data.attestation_object.auth_data.counter,
+                policy,
                 data.attestation_object.auth_data.user_verified,
                 &data.attestation_object.att_stmt,
                 data.attestation_object.auth_data_bytes,
@@ -429,6 +430,7 @@ impl<T> Webauthn<T> {
             AttestationFormat::TPM => verify_tpm_attestation(
                 acd,
                 data.attestation_object.auth_data.counter,
+                policy,
                 data.attestation_object.auth_data.user_verified,
                 &data.attestation_object.att_stmt,
                 data.attestation_object.auth_data_bytes,
@@ -437,6 +439,7 @@ impl<T> Webauthn<T> {
             AttestationFormat::AppleAnonymous => verify_apple_anonymous_attestation(
                 acd,
                 data.attestation_object.auth_data.counter,
+                policy,
                 data.attestation_object.auth_data.user_verified,
                 &data.attestation_object.att_stmt,
                 data.attestation_object.auth_data_bytes,
@@ -445,12 +448,19 @@ impl<T> Webauthn<T> {
             AttestationFormat::None => verify_none_attestation(
                 acd,
                 data.attestation_object.auth_data.counter,
+                policy,
                 data.attestation_object.auth_data.user_verified,
             ),
             _ => {
                 if self.config.ignore_unsupported_attestation_formats() {
                     let credential_public_key = COSEKey::try_from(&acd.credential_pk)?;
-                    Ok(AttestationType::None(Credential::new(acd, credential_public_key, data.attestation_object.auth_data.counter, false)))
+                    Ok(AttestationType::None(Credential::new(
+                        acd,
+                        credential_public_key,
+                        data.attestation_object.auth_data.counter,
+                        policy,
+                        data.attestation_object.auth_data.user_verified,
+                    )))
                 } else {
                     // No other types are currently implemented
                     Err(WebauthnError::AttestationNotSupported)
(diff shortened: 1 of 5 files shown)


Firstyear:
Hmmmm I'm not sure I like it. I have a few questions and comments.

First, you mentioned that this change was causing you some problems? What was the problem you were hitting? 

I think what concerns me about this approach is changing the policy from underneath the user without them knowing it changed. I don't really like "silent" changes aka surprising behaviour. I'd rather it error and fail and make noise. 

A major piece of the problem is that browsers can only apply one userverificationpolicy to the set of attached authenticators, and that policy leads to user experience inconsistencies.

The standard doesn't say it, but it assumes either there is only one authenticator in use (externally pre selected via other knowledge) or that all authenticators are consistent. 

But because it's not said, people can use it in a way that's *not* consistent or positive to a user experience, which is why I made these changes that deviate.

So it would be good to at least in any change or solution here, understand what happens in a deployment, when there are different classes of authenticators that are homogenous or mixed. For example in your current solution, because one authenticator is "required" then all become "required". But I may not *want* my yubikey to have a pin, and that causes inconsistency because the yubikey may not need a pin later. Having a deterministic and consistent behaviour is what I'm aiming for here.

To put it differently, I'm trying to make UV a property oof the *authenticator* where the standard makes it a part of the *individual authentication session* and its up to the "holder" to use it correctly. I'd rather have a library that can't be held wrong. 

Anyway, as mentioned I would like to know what was the issue you were hitting and what you are trying to achieve since that cuold help to understand how we can make this better :)

agrinman:
Yep. So the issue is that I might have a user that has let's say 2 authenticators (one of which is apple platform, the other is a yubikey). In this case I have two conflicting .verified settings (true, false) that I cannot control. I'd like to authenticate either of them (I don't care which the user chooses) and I also don't care in this case if the UV is true or false. I think this is a pretty typical setup and currently the library doesn't allow this operation.

agrinman:
So there's a two part change: first I give the developer the choice of ignoring the verification of the device in the policy decision and second I don't error if there's inconsistencies (I just choose the safest option).

Firstyear:
AAAAAAhhhhhhhhh I see. The issue is that in *registration* because we set verified on the token we try to enforce it later assuming that the supplied *policy* was also that the token should be verified.

agrinman:
Yeah I guess the workaround (which is what I'm currently doing) is hardcoding .verified to false.

agrinman:
But then I don't get the UV if the device can provide one (which could be useful for other kinds of downstream security heuristics)

Firstyear:
I agree that it's a genuine problem though. I wonder if the change should be in how we register the credential then? Really what we want to express is a *policy* about if the credential should or should not be verified, and then to maintain that consistently. So perhaps we need to store the credential verified differently during registration? 

Perhaps that if in the registration we store the policy that was requested at registration time? Then we use that to make the policy decisions rather than requesting based on the presence of the UV bit. 

So if we req discouraged, then we allow the credential to be used with discouraged even if uv was true.
If we req preferred and uv is false, we map to discouraged as a policy for that device since we can't ask it to auth later.
If we req preferred and uv is true, we may to required to keep the device providing verification in the future.
If we req required, then we store that it was created with required.

Then we know both:

* can the device do verification?
* how was it verified at registration?

So perhaps an extra boolean on the Credential type that says "verification_required". Then we enforce the policy decision in auth based on that setting, and we can then have an extra setting that says "if verification_required is False, but the credential is verified, assert still verified".


How does that sound? I think it would resolve the use case you are hitting here, and still satisfy the desire to bind UV to the credential itself and keep the consistent UV policy that I want?

agrinman:
Yep I like that. 

I'd still allow ::Preferred in the case that a library user wants to record/use that UV=1 somewhere beyond the authentication decision step.

Firstyear:
Okay, in the case that you want to over-ride that policy, how about another "generate_challenge_authenticate" but "generate_challenge_authenticate_single" and it only accepts a single credential rather than a Vec of credentials?  Internally we can just make it a Vec of len 1, but that way we know it's guaranteed that it's *only* to that one credential and we avoid the "multiple crredential confusion" issues.

That would also allow a credential to be targetted for a single *downgrade* to ignore verification also. 

Thoughts?

agrinman:
Yep, I think this makes sense. Let me update the PR.

src/core.rs
```diff
@@ -672,21 +682,34 @@ impl<T> Webauthn<T> {
...
     where
         T: WebauthnConfig,
     {
         let chal = self.generate_challenge();
 
-        let mut policy = policy.unwrap_or(UserVerificationPolicy::Preferred_DO_NOT_USE);
-
-        let any_creds_are_verified = creds.iter().fold(false, |cur, next| cur || next.verified );
+        // the credential is verified and was registered with verification required
+        let need_user_verification = creds.iter().fold(false, |cur, next| {
```
Firstyear, line 693: Still don't like this section. I think in the suggestion I made we should have two interfaces.

For generate_challenge_authenticate() it takes a Vec of credentials, and *no* policy. It has the same consistent vs inconsistent error with mixed type.

We should also have a second interface, generate_challenge_authenticate_single, which takes a single credential, and allows the policy to be over-ridden. 

That way we satisfy both your needs to have the policy changeable for a single credential (meaning the implementation has an out-of-band ability to indicate what credential should be used in this authentication, inline with the unwritten intent of the spec), and my desire that in "mixed" scenarios the policy is consistent and enforced with out "changing" the policy underneath. 

Otherwise I think all the rest looks really good!

Firstyear (changes requested):

src/core.rs
agrinman: I still think there's too much inner logic that will easily be mis-used. For example, I just ran into a 3rd case: an empty list of credentials (i.e. a user-less operation mode). In this case it **should** take a policy.

Probably you're right that it's best to have separate authenticate methods for each type of use case. There should be *two* methods that take multiple credentials -- one that takes a policy and one that doesn't. If it doesn't take a policy then it infers from the credentials existing values.

Firstyear: What would be the use case with an empty credential list? I seem to recall there is a part of the spec where you don't send a list of "allowed creds" but the server still needs the credentials available to understand how to do the authenticaiton. So there would have to be a seperate case here.

But yes, different auth methods I think is the right way here because that way the caller has a strongly typed interface, and we can document each one with exactly when and how you use it.

So we probably need another one for when the credential list is empty and we issue a challenge, but we feed the credential list seperately ....

PR #50, opened 2021-02-22
Fix uv error

agrinman:
While the larger UV issue is more complicated, I think this temporary patch is needed to not have strange behavior in the user-less operation mode.

- [x ] cargo fmt has been run
- [x ] cargo test has been run and passes
- [ x] documentation has been updated with relevant examples (if relevant)


src/core.rs
```diff
@@ -41,13 +41,18 @@ pub struct RegistrationState {
...
-    // Allowed credentials?
-    // username: UserId,
     credentials: Vec<Credential>,
     policy: UserVerificationPolicy,
     challenge: Base64UrlSafeData,
 }
 
+impl AuthenticationState {
+    /// set which credentials the user is allowed to authenticate with
+    pub fn set_allowed_credentials(&mut self, credentials: Vec<Credential>) {
```
Firstyear, line 51: I think I'd probably do this as an option to `generate_challenge_authenticate` where there is an Option<Vec<Credential>> for "credentials in this auth that we *dont* advertise in allowcredentials. I'm not sure what to name it though, but I think that's easier than trying to get people to update the Authentication state with  this knowledge. It would also potentially allow a mix of discoverable/non-discoverable creds in a single authenticatino since we can merge that list. 

Alternately, we could do this by having a generate_challenge_authenticate_<something> that advertises nothing. 

I just wonder if there are edge cases we'll hit trying to mix discoverable vs not in one auth, or if we want special interfaces here ....

Firstyear: I guess another factor is given that during registration we can request discoverable vs not discoverable, is that a factor of the `Credential` we need to understand so we know how to advertise them in the authenticate step? Perhpas that could be used as a filter in what we send in the allow credentials. But I think in authenticate, the implication of the standard is that when you want to use a non-discoverable (resident) key you send an empty credential set in challenge authenticate ....


What do you think?

Firstyear: Sorry for the spam. I think instead of a code fix, let's actually write up a design doc, where we model all the possible states and how we need to implement them, then we make the code to suit. That way we know exactly what combinations, use cases, and guarantees from the spec we have to uphold. Is that a reasonable suggestion?

agrinman: Yeah I agree...definitely needed :)

Firstyear: Okay. I'll start that today, and then you can help review and add your use cases too? I'll PR it to this repo shortly. Hows that sound mate? Really appreciate your patience and the conversation here, I think it really will help us to do the best thing for the library.

Firstyear: https://github.com/kanidm/webauthn-rs/pull/51

PR #51, opened 2021-02-22
Add doc re auth use cases

Firstyear:
@agrinman 

- [ ] cargo fmt has been run

### DIFF
--- a/designs/authentication-use-cases.md
+++ b/designs/authentication-use-cases.md
@@ -12,7 +12,9 @@ to ensure it can never be used/held incorrectly.
 
 ## Key Terms
 
-* Non-discoverable credential - https://www.w3.org/TR/webauthn-2/#credential-id
+### Non-discoverable credential
+
+[credential-id w3c](https://www.w3.org/TR/webauthn-2/#credential-id)
 
 This is the "common" type of credential that is used. These are commonly implemented with a scheme
 known as "key-wrapped-key". This is where the authenticator has a single private key, and encrypts
@@ -21,13 +23,17 @@ a credential with that key. The credential ID is the encrypted credential.
 For the authenticator to then operate, it must be presented with the credential id, that the authenticator
 decrypts and then uses in the authentication ceremony.
 
-* Discoverable Credentials - https://www.w3.org/TR/webauthn-2/#client-side-discoverable-credential
+### Discoverable Credentials
+
+[discoverable credential w3c](https://www.w3.org/TR/webauthn-2/#client-side-discoverable-credential)
 
 A discoverable credential is previously called a resident key. This is when the *client* holds the
 ability to find credentials locally rather than requiring them to be supplied in the allowedCredentials
 set in authentication.
 
-* User Verification - https://www.w3.org/TR/webauthn-2/#user-verification
+### User Verification
+
+[user verification w3c](https://www.w3.org/TR/webauthn-2/#user-verification)
 
 The process of supplying extra or supplementary authentication to the authenticator. This improves
 the authentication from "there is a person present" to "this specific owner of the authenticator" is
@@ -36,6 +42,16 @@ the device is a SFA, and other authentication is required to compose a complete 
 
 At registration we store the policy and verification that was supplied in the internal Credential
 type so that later decisions can be made for security policy.
+
+### External Knowledege
+
+This is where using elements from the user interaction, workflow, or a device fingerprint, cookie
+or other, the client or RP can make descions about which credential or policy to use before
+the initial webauthn challenge is sent.
+
+This work flow could be a user login page that requests "how" the user wishes to authenticate,
+which device they want to use. It could also be based on the device ID and understanding that
+the authenticator is part of the device, so we should prefer it's use.
 
 ## Authentication Work Flows
 
@@ -50,7 +66,7 @@ In the case all of these credentials are discouraged, we still assert that the U
 the returned authentication as it *is* valid for UV=true even when discouraged and we store that
 in the credential.
 
-* Example
+#### Example
 
 A laptop where the user has multiple UV=false yubikeys that may be used as a authentication factor.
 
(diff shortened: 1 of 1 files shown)


Firstyear:
Sorry to be a pain @agrinman would be great for you to review this so we can finish up fixing these use cases re user verification. :)

designs/authentication-use-cases.md
```diff
@@ -0,0 +1,96 @@
...
+
+In the case all of these credentials are discouraged, we still assert that the UV bit matches
+the returned authentication as it *is* valid for UV=true even when discouraged and we store that
+in the credential.
+
+* Example
+
+A laptop where the user has multiple UV=false yubikeys that may be used as a authentication factor.
+
+A user indicates they want to authenticate with their touchid as a self contained MFA. The user owns
```
agrinman, line 57: in this case it could be a mix of credentials: yubikeys + touchid that the RP would want to accept right?

designs/authentication-use-cases.md
```diff
@@ -0,0 +1,96 @@
...
+
+A longer term cookie on the device, or the device finger print indicates to the server that we can
+use a discoverable credential from that device.
+
+## Work flows that may not be relevant?
+
+### Homogenous Discoverable Credentials
+
+As a discoverable credential we already required external knowledge of what credential will be used
+so does it become needed to have many discoverable credentials in one transaction?
```
agrinman, line 91: possibly. touchid on my iphone + mac. the rp shouldn't have to know which device the user is on but wants to accept both. am i missing something here?

agrinman (commented):
Sorry for the delay on this! After taking a break from looking at this I am coming back to some of my original thinking so  sorry if this repeating some old material. 

Is it not the most typical case that an RP wants to send down all available credentials to maximize the opportunity that the auth will work: yubikeys, resident keys (ios, macbook, etc)? In this case the required UV could all be mixed and the RP would not force a particular preference. Should we optimize for the common case and provide mechanisms to enable the other methods?

designs/authentication-use-cases.md
Firstyear: Yep, absolutely. So yubikey + touchid, and the touchid could be uv=true at registration, but we use the uv policy at registration for consistency I think. So if we register with uv=discouraged, but we get back uv=true from the device, then we still assert that holds true. 

I think maybe i'll update this section with some more detail examples of the registration through to auth flow though. I wonder if this could make uv=preferred at registration, and then we just check each credentials uv=true/false would then be viable for these situations.

designs/authentication-use-cases.md
Firstyear: Yeah, I thought about that possibility too. From my understanding, the intent is for a discoverable credential to be used, that the server has "pre-knowledge" about the client so it knows "what to expect" in response. This could be a device id, a cookie, fingerprint ...

But I can certainly also see how without the external knowledge, we'd need to accept "well it could be any device now". So that's why it's a "question" if we need this or not? Again, not opposed, but it's good to just have those work flows, and I think we can expand this based on your suggestion, and the fact we make not have pre-knowledge.

Saying this, how would this interact with non-discoverable credentials? How would we have a situation where we say and present a UI that says "okay, what device did you want to use?" I think that may be my concern. We can't mix discoverable and non-discoverable in one transaction, so that's where the "external knoweldge" is needed, and once you have that external knowledge, you may as well select the credential matching that device?

Firstyear (commented):
> Sorry for the delay on this! After taking a break from looking at this I am coming back to some of my original thinking so sorry if this repeating some old material.

No problem at all, not repeating. It's good to be clear :) 

> 
> Is it not the most typical case that an RP wants to send down all available credentials to maximize the opportunity that the auth will work: yubikeys, resident keys (ios, macbook, etc)? In this case the required UV could all be mixed and the RP would not force a particular preference. Should we optimize for the common case and provide mechanisms to enable the other methods?

The issue is that we can't send both discoverable and non-discoverable credentials in one txn I think. the client only activates discoverable credentials if the allowCredentials set is empty. So we need the external knowledge or a hint from the client about "what" we should set into allow credentials. 

And yes the UV can be mixed here, but are we talking that the UV *policy* is mixed? Or the UV bits we get back? 

I think we can check the UV-bits based on the used credential ID, regardless of the policy, but I think the "intent" of the standards authors which isn't always clear, is that the client pre-selects through some ui or other workflow *before* the RP sends the challenge, so there shuold already be a way for the RP to know "okay, we probably will see this device used". 

Still I may extend this to be a bit clearer about some possibilities today. Especially the possible choices and states between UV policy at registration and the sign in to be sure that we can have a consistent experience.

Firstyear:
Updated with more details about some use cases.

designs/authentication-use-cases.md
```diff
@@ -0,0 +1,192 @@
...
+
+| UV Policy | UV boolean returned | Valid UV Policies for Auth | UV boolean always checked in auth |
+| --------- | ------------------- | -------------------------- | --------------------------------- |
+| discouraged | false             | discouraged                | No - device can not do UV |
+| discouraged | true              | discouraged, required      | Yes - device always does UV |
+| preferred   | false             | discouraged                | No - device can not do UV |
+| preferred   | true              | discouraged, required      | No - device may not always send UV in discouraged mode |
+| required    | false             | -                          | No - device can not be used |
+| required    | true              | required                   | Yes - device should always performs UV |
+
```
Firstyear, line 90: It may also be worth making clear why preferred isn't a valid policy for auth (credentials registered with discouraged can be forced to use UV=true even though it wasn't true at registration, breaking user expectations, and we don't have the internal knowledge to know it's possible )

agrinman:
> The issue is that we can't send both discoverable and non-discoverable credentials in one txn I think. the client only activates discoverable credentials if the allowCredentials set is empty. So we need the external knowledge or a hint from the client about "what" we should set into allow credentials.
> 

Why not? I think the authenticator can make that call however it likes right? There's no limitation that the list has to be empty to accept a discoverable credential -- or am I missing something?

Firstyear:
> > The issue is that we can't send both discoverable and non-discoverable credentials in one txn I think. the client only activates discoverable credentials if the allowCredentials set is empty. So we need the external knowledge or a hint from the client about "what" we should set into allow credentials.
> 
> Why not? I think the authenticator can make that call however it likes right? There's no limitation that the list has to be empty to accept a discoverable credential -- or am I missing something?

There are two issues here :) it's no problem, it's a big standard so easy to miss stuff.

https://www.w3.org/TR/webauthn-2/#sctn-verifying-assertion

7.2 step 5:

```
If options.allowCredentials is not empty, verify that credential.id identifies one of the public key credentials listed in options.allowCredentials.
```

This means if we have non-discoverable credentials, we can't mix them with discoverable ones, since the moment we offer an allowCredentials list, the credential used MUST be part of that set.

In the inverse, a standards compliant client:

https://www.w3.org/TR/webauthn-2/#sctn-op-get-assertion

6.3.3 step 3 and 4

```
3 If allowCredentialDescriptorList was supplied, then for each descriptor of allowCredentialDescriptorList:

    Let credSource be the result of looking up descriptor.id in this authenticator.

    If credSource is not null, append it to credentialOptions.

4 Otherwise (allowCredentialDescriptorList was not supplied), for each key → credSource of this authenticator’s credentials map, append credSource to credentialOptions.
```

Which again, means that if we send an allow credential list the client will not allow us to use discoverable credentials. Alternately if we DONT send the list we can ONLY use discoverable ones. 

That's why we have to have the separation. The "intent" of the standards authors here is you use external knowledge to select ahead of time before you send the auth challenge. I don't particularly agree, I think I agree with you better, why can't we just send it all. But this is the standard we have, no the standard we want :)

agrinman:
Yep ok got it. So if you send cred ids down, then it must be non-discoverable.

Back to UV though, it is still possible that non-discoverable creds have a mixture of UV bits -- right?

Firstyear:
> Yep ok got it. So if you send cred ids down, then it must be non-discoverable.
> 
> Back to UV though, it is still possible that several non-discoverable creds have a mixture of UV bits -- right?

Yep, exactly the same. And with just the same issues with policy. But I think in a way at least it's simpler that the discoverable/non-discoverable aren't mixed together at least.

So that's why there is so much "emphasis" on "external knowledge" before the registration/authentication begins re the uv policy and what credentials are being used. Sadly the standard isn't completely clear that there is extra steps you need to do before hand.

Firstyear:
Just was made aware that another group found these issues before I did, and it was given a CVE in NC and azure AD was affected. https://hwsecurity.dev/2020/08/webauthn-pin-bypass/

Firstyear:
@agrinman Hey mate, sorry to disturb, did you get some more time to cehck this? Would be good to have it merged so we can do some code and unblock you :)

agrinman:
Sorry been quite busy recently...yes I think this is good to merge. I think the design of the api should at least make it easy for the common case where the RP will accept any credential that a client can offer [this will be the most common case I bet] but I don't think that contradicts anything in here :)

Firstyear:
Awesome, I think from here we can start to make the apis to match up to this. Appreciate all your patience and input!

PR #56, opened 2021-03-30
Preliminary addition of `credProtect` support

ericmarkmartin:
This pull request allows sending of the `credProtect` extension down to the client and deserializing the `credProtect` outputs the client gives back.


examples/tide/actors.rs
```diff
@@ -132,21 +132,24 @@ impl WebauthnActor {
...
-                    Ok(())
-                }
-                None => {
-                    // Invalid state but not end of world ...
-                    Err(())
-                }
-            };
-            ()
-        });
+        let r = self
```
ericmarkmartin, line 135: Surprised `cargo-fmt` didn't hit this already

src/core.rs
```diff
@@ -371,10 +373,12 @@ impl<T> Webauthn<T> {
         // TODO: Today we send NO EXTENSIONS, so we'll never have a case where the extensions
         // are present! But because extensions are possible from the config we WILL need to manage
         // this situation eventually!!!
+
         match &data.attestation_object.auth_data.extensions {
             Some(_ex) => {
                 // We don't know how to handle client extensions yet!!!
-                return Err(WebauthnError::InvalidExtensions);
+                println!("ext: {:?}", _ex);
+                // return Err(WebauthnError::InvalidExtensions);
```
ericmarkmartin, line 381: Level 2 doesn't want us to reject all unsolicited or invalid extensions by default anymore. Should we be doing anything else here (left the print and comment as a reminder to myself).

src/core.rs
```diff
@@ -674,7 +683,7 @@ impl<T> Webauthn<T> {
     pub fn generate_challenge_authenticate_extensions(
         &self,
         creds: Vec<Credential>,
-        extensions: Option<JSONExtensions>
+        extensions: Option<JSONExtensions>,
```
ericmarkmartin, line 686: Should this also be of the new extensions input type? It wasn't super clear to me if there should be different extension types for different parts of the protocol.

src/core.rs
```diff
@@ -911,7 +920,7 @@ pub trait WebauthnConfig {
     /// 15. If validation is successful, obtain a list of acceptable trust anchors (attestation
     /// root certificates or ECDAA-Issuer public keys) for that attestation type and attestation
     /// statement format fmt, from a trusted source or from policy. For example, the FIDO Metadata
-    /// Service [FIDOMetadataService] provides one way to obtain such information, using the
+    /// Service \[FIDOMetadataService\] provides one way to obtain such information, using the
```
ericmarkmartin, line 923: Doc generator yelled at me without escaping these

src/core.rs
```diff
@@ -937,6 +946,23 @@ pub trait WebauthnConfig {
...
+    ///
+    /// 17: Verify that the values of the client extension outputs in
+    /// clientExtensionResults and the authenticator extension outputs in the
+    /// extensions in authData are as expected, considering the client extension
+    /// input values that were given in options.extensions and any specific
+    /// policy of the Relying Party regarding unsolicited extensions, i.e.,
+    /// those that were not specified as part of options.extensions. In the
+    /// general case, the meaning of "are as expected" is specific to the
+    /// Relying Party and which extensions are in use.
+    fn reject_unsolicited_extensions() -> bool {
```
ericmarkmartin, line 963: Haven't done anything with this—but curious what you think of it as a matter of API

src/proto.rs
```diff
@@ -340,6 +341,95 @@ pub struct AllowCredentials {
...
+    UserVerificationRequired = 0x3,
+}
+
+impl TryFrom<u8> for CredentialProtectionPolicy {
+    type Error = &'static str;
+
+    fn try_from(v: u8) -> Result<Self, Self::Error> {
+        use CredentialProtectionPolicy::*;
+        match v {
+            0x1 => Ok(UserVerificationOptional),
```
ericmarkmartin, line 369: I need to add a test here because this conversion isn't tightly bound to the actual enum values.

src/proto.rs
```diff
@@ -540,21 +630,21 @@ impl Into<web_sys::CredentialRequestOptions> for RequestChallengeResponse {
                     &obj,
                     &JsValue::from("type"),
                     &JsValue::from_str(ac.type_.as_str()),
-                );
+                ).unwrap();
```
ericmarkmartin, line 633: Got warnings about unused result here (and below)

src/proto.rs
```diff
@@ -652,43 +743,43 @@ pub struct AuthenticatorData {
...
-            .map(|v| cbor_try_bytes!(v).ok())
-            .flatten()
-            .map(|v| v.as_slice())
-    }
-}
+    pub(crate) extensions: Option<AuthenticatorExtensionsClientOutputs>,
+    // pub(crate) extensions: Option<serde_cbor::Value>,
+}
+
+// impl AuthenticatorData {
```
ericmarkmartin, line 750: Not sure that we need this anymore, but wasn't sure about its original motivation

src/proto.rs
```diff
@@ -652,43 +743,43 @@ pub struct AuthenticatorData {
...
 
-    // Now re-encode it to find the length ... yuk.
-    let encoded = serde_cbor::to_vec(&v)
-        .map_err(|_| nom::Err::Failure(nom::Context::Code(i, nom::ErrorKind::Custom(2))))?;
-    // Finally we know the cred len
-    let cred_len = encoded.len();
+    let len = deserializer.byte_offset();
 
-    Ok((&i[cred_len..], v))
+    Ok((&i[len..], v))
```
ericmarkmartin, line 772: Renamed from `cred_len` as it no longer only parser credentials

src/proto.rs
```diff
@@ -1631,4 +1723,20 @@ mod tests {
         let tpmt_sig = TpmtSignature::try_from(data.as_slice()).unwrap();
         println!("{:?}", tpmt_sig);
     }
+
+    #[test]
```
ericmarkmartin, line 1727: I'm planning on adding some more tests--this is just a start

src/proto.rs
```diff
@@ -652,43 +743,22 @@ pub struct AuthenticatorData {
     /// The optional attestation.
     pub(crate) acd: Option<AttestedCredentialData>,
     /// Extensions supplied by the device.
-    pub(crate) extensions: Option<serde_cbor::Value>,
-}
-
-impl AuthenticatorData {
```
ericmarkmartin, line 658: I don't think we need this anymore, but wasn't sure what it's original motivation was

examples/tide/actors.rs
Firstyear: I am testing with 1.46, and cargo fmt changes with different versions.

src/proto.rs
```diff
@@ -340,6 +341,95 @@ pub struct AllowCredentials {
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub enforce_credential_protection_policy: Option<bool>,
+}
+
+/// Extension option inputs for [PublicKeyCredentialCreationOptions]
+#[derive(Debug, Serialize, Clone, Deserialize)]
+pub struct AuthenticatorExtensionsClientInputs {
+    /// The credProtect extension
+    #[serde(flatten)]
+    pub cred_protect: Option<CredProtectExtensionClientInput>,
```
Firstyear, line 422: I think this should be Option<Box<Cred...>> because else this struct could get pretty large. See what clippy says I guess ....

src/proto.rs
```diff
@@ -340,6 +341,95 @@ pub struct AllowCredentials {
...
+    /// Whether it is better for the authenticator to fail to create a
+    /// credential rather than ignore the protection policy
+    /// If no value is provided, the client treats it as `false`.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub enforce_credential_protection_policy: Option<bool>,
+}
+
+/// Extension option inputs for [PublicKeyCredentialCreationOptions]
+#[derive(Debug, Serialize, Clone, Deserialize)]
+pub struct AuthenticatorExtensionsClientInputs {
```
Firstyear, line 419: This type will likely need a AuthenticatorExtensionsBuilder to make it easier for people to create this in their implementations.

src/core.rs
Firstyear: I think it should be different to indicate that the extensions may only be valid in registration or assertion contexts yes.

Firstyear (changes requested):
Just waiting on the other tests I think but it looks good!

src/core.rs
ericmarkmartin: Right--are you okay with escaping it like this?

src/core.rs
ericmarkmartin: And if we do know the extension?

src/proto.rs
ericmarkmartin: I'll see what it says. It's also worth noting that we could remove the option at the top level here and just have extensions implement default. 

I chose the option instead for size and potential ease of use, but curious what you think.

src/core.rs
ericmarkmartin: We haven't implemented extensions for authentication, so are we good leaving this as `Option<JSONExtensions>`

src/proto.rs
ericmarkmartin: Clippy doesn't seem to mind.

src/core.rs
Firstyear: The key word is unsolicited. If we are able to parse and handle it, even if unsolicited, we can't reject it.

src/core.rs
Firstyear: we should rename it to be something clearer like `Option<RequestAssertionExtension>`. At the same time we probably want to make a similar name change for `Option<RequestRegistrationExtension>`.

src/core.rs
Firstyear: Yep

src/proto.rs
Firstyear: Cool, in that case lets leave it for now.

Firstyear (changes requested):

src/core.rs
ericmarkmartin: What is `Option<RequestRegistrationExtension>`?

Firstyear: Rename `extensions: Option<AuthenticatorExtensionsClientInputs>,` to `Option<RequestRegistrationExtension>` is what I mean.

src/core.rs
ericmarkmartin: Okay so ignore unknown extensions, and parse known ones. I can also add a field in the extensions output struct that picks up any other extension values so webauthn-rs clients can handle stuff we don't yet support, though perhaps you would prefer people not leaning on that

src/core.rs
ericmarkmartin: Okay. Do you think there might be a better word than `Request`? I worry a bit because although from the RP's (and thus `webauthn-rs`'s) perspective this is like a "request" we send to the client, it is technically a "response" in the world of http(s).

ericmarkmartin: Perhaps `RegistrationExtensionInputs`s? I'll also note in case it wasn't clear that I got the name `AuthenticatorExtensionsClientInputs` from the spec.

ericmarkmartin: Also not to get too fussy but could singularizing extensions to extension perhaps confuse users who think that the struct will be the input data for a single extension?

Firstyear: I'd say that the terminology used in the Webauthn spec is "a hot mess" at best, and generally incomprehensible to most people (sometimes even leading to security issues but you know ... ). So sometimes it can be best to avoid it, and to use human accessible names, and just put in our code documentation that "this is the X term from the webauthn spec".

Anyway, the reason I said Option<RequestRegistrationExtensions> is because "we are requesting these extensions to be added by the client when it signs this registration". So I think from that direction the term Request here makes sense. 

Alternately, just "RegistrationExtensions" would work too.

Also yes, it should be plural, I think I mistyped.

ericmarkmartin: Ah, I see. Then what would you suggest naming the outputs struct? `RegistrationExtensionsOutput`?

Firstyear: SignedRegistrationExtensions?
RegistrationExtensionsResponse?
FlargbolAbloby?

ericmarkmartin: I like that last one XD

Firstyear: Yolo it to production then :D 

More seriously, I think SignedRegistrationExtensions is the best here, because it indicates clearly that it's different, and that they can be trusted by the RP.

ericmarkmartin: Sounds good—I guess we'd just need to be careful that we never spit that struct back out at a user until we've actually validated the signature

Firstyear: We would only return this as part of attested credential data after a full and complete verification through our code anyway, so I think that this is fine.

Firstyear: Also, since this is still 0.3 alpha, worst case we can rename it anyway because we can still make breaking changes. 

I'd actually like to merge this today as I want to work on some other changes of the api today/tomorrow, so did you want to do the tests in a second pr?

ericmarkmartin: Sounds good. I'll push the renaming changes and PR the tests later

PR #60, opened 2021-04-16
Authentication Verification Refactor

Firstyear:
Supercedes #49 - this implements the start of the design as discussed in https://github.com/kanidm/webauthn-rs/blob/master/designs/authentication-use-cases.md

This allows:

* Multiple credentials, with a derive UV policy and checked for consistency.
* A single credential, which can have a policy over-ride applied.

- [ - ] documentation has been updated with relevant examples (if relevant)

### DIFF
--- a/src/core.rs
+++ b/src/core.rs
@@ -600,7 +600,7 @@ impl<T> Webauthn<T> {
             }
             (UserVerificationPolicy::Discouraged, UserVerificationPolicy::Discouraged) => {
                 // If this token always verifies, even under discouraged, we can enforce that.
-                if cred.verified && self.get_require_uv_consistency() {
+                if cred.verified && self.config.get_require_uv_consistency() {
                     // This token always sends UV, so we enforce that.
                     if !data.authenticator_data.user_verified {
                         log::debug!("Token registered UV=true with DC, enforcing UV policy.");
--- a/src/core.rs
+++ b/src/core.rs
@@ -726,10 +726,9 @@ impl<T> Webauthn<T> {
     where
         T: WebauthnConfig,
     {
-        let (verified, unverified): (Vec<Credential>, Vec<Credential>) =
-            creds.into_iter().partition(|cred| {
-                cred.registration_policy == UserVerificationPolicy::Required
-            });
+        let (verified, unverified): (Vec<Credential>, Vec<Credential>) = creds
+            .into_iter()
+            .partition(|cred| cred.registration_policy == UserVerificationPolicy::Required);
 
         match (verified.len(), unverified.len()) {
             (_, 0) => self.generate_challenge_authenticate_inner(
@@ -1943,7 +1942,8 @@ mod tests {
         // Ensure we get a bad result.
 
         assert!(
-            wan.generate_challenge_authenticate(creds.clone()).unwrap_err()
+            wan.generate_challenge_authenticate(creds.clone())
+                .unwrap_err()
                 == WebauthnError::InconsistentUserVerificationPolicy
         );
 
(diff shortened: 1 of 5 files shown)


src/core.rs
```diff
@@ -114,11 +114,16 @@ impl<T> Webauthn<T> {
     pub fn generate_challenge_register(
         &self,
         user_name: &str,
-        policy: Option<UserVerificationPolicy>,
+        user_verification_required: bool,
```
agrinman, line 117: any reason we  don't  just pass the policy directly here?

src/core.rs
```diff
@@ -574,14 +583,30 @@ impl<T> Webauthn<T> {
...
                     return Err(WebauthnError::UserNotVerified);
                 }
             }
-            UserVerificationPolicy::Preferred_DO_NOT_USE | UserVerificationPolicy::Discouraged => {}
-        };
+            (UserVerificationPolicy::Discouraged, UserVerificationPolicy::Discouraged) => {
+                // If this token always verifies, even under discouraged, we can enforce that.
+                if cred.verified {
+                    // This token always sends UV, so we enforce that.
+                    if !data.authenticator_data.user_verified {
```
agrinman, line 600: I think this might cause some false positive errors. perhaps we want to control this with a config?
i.e: `if cred.verified && self.config.get_require_uv_consistency()`?

src/core.rs
```diff
@@ -639,6 +664,26 @@ impl<T> Webauthn<T> {
...
+    ///
+    /// NOTE: Over-riding the UserVerificationPolicy may have SECURITY consequences. You should
+    /// understand how this interacts with the single credential in use, and how that may impact
+    /// your system security.
+    ///
+    /// If in doubt, do NOT use this function!
+    pub fn generate_challenge_authenticate_credential(
+        &self,
+        cred: Credential,
+        policy: UserVerificationPolicy,
```
agrinman, line 678: why would we allow overriding the policy? this seems like an easy footgun

agrinman (commented):
I think it's close, just a couple of comments above for discussion.

One other suggestion: let's add tests for our UV test cases because it's  very not obvious :) I.e. tests that check all the combinations for setting `credential.verified` and `policy` and ensuring that `register_internal` and `verify_internal` work as intended.

src/core.rs
Firstyear: This is the "basic" register function, so I thought to keep it as UV=true/false for discouraged/required to make it simpler to hold. The policy can be directly used in the options version.

src/core.rs
Firstyear: I seem to remember you requested the ability to override the UV policy for a singular credential for some use cases :) That's why I added it because I think for a single credential, it's still a footgun, but it's not an automatic foot chainsaw like policy on a set of credentials is.

src/core.rs
Firstyear: Yep, this is a good point. I'll add that switch now.

Firstyear:
Just need to add the tests later of the various combinations. :)

Firstyear:
Updated with some tests of the UV consistency options.

ericmarkmartin (approved):

Firstyear:
I'll let @agrinman have another look before I merge :)

src/core.rs
agrinman: Perhaps I misread, but I thought I saw that in `verify_credential_internal` it would error if mismatched.
I think that the right way to "override" it is to just change it on the credential itself.

agrinman (approved):

src/core.rs
agrinman: nvm ignore this!

Firstyear: Yeah, when you set the override we don't do the policy consistency check. The only thing that *could* occur, is that the credential will verify under discouraged which we have a seperate flag for checking :)

Issue #61, opened 2021-04-18
Take more arguments by reference when possible


hgzimmerman:
There are a few cases in this crate's API where a function takes an argument by value, when it could easily take it by reference instead, without additional cloning.

The effect of this is that code that uses the library often needs to clone the items with which it calls the functions defined in this crate. Ideally, this could be minimized or eliminated.

This, of course, would necessitate some breaking changes.

PR #62, opened 2021-04-18
Refactor functions to take more arguments by reference

hgzimmerman:
Implements #61.

* Adds `ChallengeRef`, to allow working with `Challenge`s in the same way that `&[T]` and `Vec<T>` are used.
  * This adds a small amount of unsafe code, but it is very mundane as far as unsafe code goes.
* `Webauthn::register_credential_internal()` takes `chal` and `exclude_credentials` by reference.
* `Webauthn::register_credential()` takes `state` by reference.
* `Webauthn::authenticate_credential()` takes `state` by reference and returns a reference to a `CredentialID`.
* `WebauthnConfig`'s methods:  `get_relying_party_name` and `get_relying_party_id` now return `&str` instead of `String`.
  * I'm not sure if this is ideal. I understand that webauthn_rs code currently needs to clone these `&str`s anyways, but the trait was asking that the implementer perform `&self -> String` which will always require a reallocation (or questionable use of interior mutability) to get ownership of a `String` to return.
  * Making the functions return a &str limits the ability of the implementer of the trait to `format!()` them if needed. If you believe that ad-hoc formatting in these functions is a common, then defining the function to return `Cow<'_, str>` should be done instead to allow either pass an unformatted str or a formatted String.

Some enums in this crate that could be `Copy` are not, I'm also wondering if this is done with the understanding that they may become not `Copy` in the future, or if they could be made to derive that marker trait.

- [x] cargo fmt has been run

### DIFF
--- a/examples/tide/actors.rs
+++ b/examples/tide/actors.rs
@@ -88,7 +88,7 @@ impl WebauthnActor {
         let r = match creds.get_mut(&username) {
             Some(ucreds) => self
                 .wan
-                .register_credential(reg, rs, |cred_id| Ok(ucreds.contains_key(cred_id)))
+                .register_credential(reg, &rs, |cred_id| Ok(ucreds.contains_key(cred_id)))
                 .map(|cred| {
                     let cred_id = cred.0.cred_id.clone();
                     ucreds.insert(cred_id, cred.0);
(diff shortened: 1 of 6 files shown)


Firstyear:
Thanks!

Issue #55, opened 2021-03-22
Support credProtect and credBlob


Firstyear:
https://fidoalliance.org/specs/fido-v2.1-rd-20210309/fido-client-to-authenticator-protocol-v2.1-rd-20210309.html#sctn-credProtect-extension

ericmarkmartin:
With #56 and #66 merged, should we close this? There's more to do, but such tasks are perhaps better captured by smaller issues like #67.

Firstyear:
Yep, I think so :)

PR #66, opened 2021-04-21
add cred blob extension

ericmarkmartin:
-------

Implements #55 (together with #56).

- [x] cargo fmt has been run

_Note: cannot yet be tested against browser due to lack of implementation_


src/proto.rs
```diff
@@ -576,22 +639,62 @@ impl RequestRegistrationExtensionsBuilder {
...
+    /// ```
+    /// # use webauthn_rs::proto::{RequestRegistrationExtensions, CredBlobSet};
+    /// let cred_blob = vec![0xde, 0xad, 0xbe, 0xef];
+    /// let extensions = RequestRegistrationExtensions::builder()
+    ///     .cred_blob(cred_blob.clone())
+    ///     .build();
+    ///
+    /// assert_eq!(extensions.cred_blob, Some(CredBlobSet::from(cred_blob)));
+    /// ```
+    pub fn cred_blob(mut self, cred_blob: Vec<u8>) -> Self {
```
Firstyear, line 659: Should this be "set_cred_blob"?

Firstyear, line 659: I'm wondering if we should length check this, and do Result<Self, Self>, where Err(Self) is the builder but without cred_blob set. 

IIRC there was a 32byte limit?

ericmarkmartin: I think implementing authenticators are required at accept at least 32 bytes of data, but technically can accept more

ericmarkmartin: This is actually one of the reasons you can get a `False` return on the way back

Firstyear: Ahhh because you requested to store something too large. Got it. Okay in that case, this is all good :)

ericmarkmartin: This is a good question. My modeling of the builder was after how `tide` does there's, in which the names of the setters are the names of the fields they set. I suppose there's a bit of ambiguity if we do `set_cred_blob`, in whether "`set`" refers to `set`ting the field in the struct or `set`ting the cred blob data for the authenticator

ericmarkmartin: If the latter, am I correct in assuming that we would also want `cred_blob` -> `get_cred_blob` for the other builder? One could also argue for `set_cred_protect`, I suppose.

Firstyear: Ahh yeah. Well, we also have just "cred_protect" here too. Honestly it's much-over-muchness, so I think this is fine :) I retract my comment :)

Firstyear:
As always, great work mate!

Issue #70, opened 2021-04-24
Use Owned Strings instead of References in Configuration


jalcine:
When attempting to implement [`WebAuthnConfig`](https://docs.rs/webauthn-rs/0.2.5/webauthn_rs/trait.WebauthnConfig.html), it has values whose return values are `&str`. This makes it difficult for my application to provide values that are instantiated from configuration.

jalcine:
I understand that `&str` is commonly used as the _argument_ type of strings but for returning, I figured that `String` would be more optimal.

Issue #63, opened 2021-04-18
Improve resident key handling and attachement


Firstyear:
These should probably be options to the registration or authentication functions given the way they work, rather than site wide settings.

Firstyear:
These are now split into options and the wrapper can handle this better. We also have credprops soon (even if it's lol).

PR #71, opened 2021-04-24
Use owned strings instead of references.

jalcine:
Implements #70.

- [x] cargo fmt has been run

### DIFF
--- a/src/core.rs
+++ b/src/core.rs
@@ -829,7 +829,7 @@ pub trait WebauthnConfig {
     /// confuse authenticators and will cause their credentials to be lost.
     ///
     /// Examples of names could be "My Awesome Site", "https://my-awesome-site.com.au"
-    fn get_relying_party_name(&self) -> &str;
+    fn get_relying_party_name(&self) -> String;
 
     /// Returns a reference to your sites origin. The origin is the URL to your site with
     /// protocol and port. This should rarely, if ever change. In production usage this
@@ -838,7 +838,7 @@ pub trait WebauthnConfig {
     /// may cause associated authenticators to lose credentials.
     ///
     /// Examples of this value could be. "https://my-site.com.au", "https://my-site.com.au:8443"
-    fn get_origin(&self) -> &str;
+    fn get_origin(&self) -> String;
 
     /// Returns the relying party id. This should never change, and is used as an id
     /// in cryptographic operations and credential scoping. This is defined as the domain name
@@ -848,7 +848,7 @@ pub trait WebauthnConfig {
     /// If changed, all associated credentials will be lost in all authenticators.
     ///
     /// Examples of this value for the site "https://my-site.com.au/auth" is "my-site.com.au"
-    fn get_relying_party_id(&self) -> &str;
+    fn get_relying_party_id(&self) -> String;
 
     /// Get the list of valid credential algorthims that this service can accept. Unless you have
     /// speific requirements around this, we advise you leave this function to the default
--- a/src/ephemeral.rs
+++ b/src/ephemeral.rs
@@ -43,18 +43,18 @@ impl std::fmt::Debug for WebauthnEphemeralConfig {
 
 impl WebauthnConfig for WebauthnEphemeralConfig {
     /// Returns the relying party name. See the trait documentation for more.
-    fn get_relying_party_name(&self) -> &str {
-        &self.rp_name
+    fn get_relying_party_name(&self) -> String {
+        self.rp_name.clone()
     }
 
     /// Returns the relying party id. See the trait documentation for more.
-    fn get_relying_party_id(&self) -> &str {
-        &self.rp_id
+    fn get_relying_party_id(&self) -> String {
+        self.rp_id.clone()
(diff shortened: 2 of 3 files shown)


Firstyear:
Curious what the use case is that makes this not work from a configuration. You should be able to create a struct that stores the String identifies of your site you need, and that's what allows you to get the &str. This is what the ephemeral config does. Why would this need to be a String type, which would imply that this value is dynamically changing?

jalcine:
> [@Firstyear]: Curious what the use case is that makes this not work from a configuration. You should be able to create a struct that stores the String identifies of your site you need, and that's what allows you to get the &str. This is what the ephemeral config does. Why would this need to be a String type, which would imply that this value is dynamically changing?

I'm working on using WebAuthn to do authentication for a self-hostable [IndieAuth](https://indieauth.spec.indieweb.org) server - think OAuth2 but you can use your URL on the Web as the ID. That requires me to be able to define things like the origin and host on the fly.

Firstyear:
@jalcine The issue is that credentials need to be bound to the origin they are registered at. To have this dynamically as you are suggesting could open the possibility a credential registered to one origin can be used on another. That's in violation of the standard and it's intent. 

@agrinman What do you think? We could make origin a property of the credential which would allow this dynamic case. We would send the origin from within the credential during the authentication to ensure it's consistent with how it was registered, and we could then make origin a function argument in registration. More over, I think maybe the "config trait" object is maybe not the best way to handle these things, so perhaps we need to rethink that design?

Firstyear:
cc @ericmarkmartin as you've been looking at this a lot lately and may have thoughts too.

agrinman:
If you're operating multiple RPs you can just have multiple implementors of the webauthn config trait. I don't think this restricts anything. The only way a reference wouldn't work is if you're dynamically generating the rpid each time...

Firstyear:
@agrinman I think that's what they are doing here. They want to dynamically associate a credential to different origin/rp_ids, or they want to have one webauthn config to serve many origins.

Regardless, Alex's response is a good point, you can many different implementations of the webauthn config type with each origins policy defined.

ericmarkmartin:
I also think @agrinman's response makes sense

I also think that as with #63 I think it's probably worth taking some time to think about how we decide which credential properties are defined on configs and which can be passed directly to registration/authentication functions, but providing users with the ability to pass the origin directly into the credential functions might be a footgun as suggested in [your comment](https://github.com/kanidm/webauthn-rs/pull/71#issuecomment-826165664).

Firstyear:
@ericmarkmartin I think we'd only pass in the origin at registration, then it would be a property of the credential and could not be an option in the future. That would bind the credential to that origin for all future authentications.

ericmarkmartin:
@jalcine, I'm curious as to how additionally complicating it would be for you to use multiple implementors of `WebauthnConfig`.

Firstyear:
I think the otherside of this today is that a credential is not bound to a specific webauthn config struct. So they could still put a credential regisitred from one origin/webauthnconfig into another for it's authentication which also would violate some guarantees.

ericmarkmartin:
> @ericmarkmartin I think we'd only pass in the origin at registration, then it would be a property of the credential and could not be an option in the future. That would bind the credential to that origin for all future authentications.

Right, but as you pointed out such a credential could then be used against other origins.

ericmarkmartin:
> I think the otherside of this today is that a credential is not bound to a specific webauthn config struct. So they could still put a credential regisitred from one origin/webauthnconfig into another for it's authentication which also would violate some guarantees.

Hmm I hadn't thought about this

jalcine:
> If you're operating multiple RPs you can just have multiple implementors of the webauthn config trait. I don't think this restricts anything. The only way a reference wouldn't work is if you're dynamically generating the rpid each time...

This is _more or less_ what happens, defining the RP ID on boot because the host to use isn't known until it's read from configuration.

Firstyear:
@jalcine So each RP only hosts one origin and you read it from a config file .... I'm not seeing the issue then?

jalcine:
> @jalcine So each RP only hosts one origin and you read it from a config file .... I'm not seeing the issue then?

I can't seem to pass a `String` to the value here. I'm afraid now that this is my lack of understanding of how to return a `&str` when it can't be borrowed (thus implying this should be a constant value but that's not applicable for my use case).

Firstyear:
Your program order should be:

```
1) Read the config file
2) Create the singular webauthn config struct that you create and have derived webauthnConfig to. 
3) Inside that webatuhn config struct, you give it the string (via clone() or whatever)
4) self.origin.as_str() and it all works. 
```

Firstyear:
And then repeat as many times as you need for each configuration.

jalcine:
Welp, that worked! Thank you @Firstyear - apologies for any inconvenience by this. I'll close the PR as well

Firstyear:
No problem, it's raised some interesting questions that I think @agrinman, @ericmarkmartin and myself should think about in how we approach this :) If you have any other questions, please let us know!

Issue #52, opened 2021-02-24
Be more tolerant to changes in CCD


Firstyear:
https://www.w3.org/TR/webauthn-2/#dictionary-client-data

This will change in the future to include crossOrigin, so we must be ready for this to occur and handle it.

ericmarkmartin:
A quick test indicates that chromium already sends crossOrigin, whereas Firefox does not

ericmarkmartin:
We could either add a field `pub cross_origin: Option<bool>` so that we can capture the value from browsers that send it, or we can add a `BTreeMap` or `HashMap` to capture unknown keys, wait until `cross_origin` is better adopted, and then add it as `cross_origin: bool`.

Firstyear:
I think doing it with Options makes more sense :)

Firstyear:
This seems to be resolved :)

PR #72, opened 2021-04-25
Update CCD to spec

ericmarkmartin:
Add `crossOrigin` field to `CollectedClientData` and capture unknown keys in an extra field.

Additionally updates some formatting left over from credBlob.

Implements #52.

- [x] cargo fmt has been run

### DIFF
--- a/src/core.rs
+++ b/src/core.rs
@@ -297,16 +297,21 @@ impl<T> Webauthn<T> {
         }
 
         // Verify that the value of C.origin matches the Relying Party's origin.
-        if !self.config.allow_cross_origin()
-            && (data.client_data_json.cross_origin.unwrap_or(false)
-                || data.client_data_json.origin != self.config.get_origin())
-        {
+        if data.client_data_json.origin != self.config.get_origin() {
             log::debug!(
                 "{} != {}",
                 data.client_data_json.origin,
                 self.config.get_origin()
             );
             return Err(WebauthnError::InvalidRPOrigin);
+        }
+
+        // ATM most browsers do not send this value, so we must default to
+        // `false`. See [WebauthnConfig::allow_cross_origin] doc-comment for
+        // more.
+        if !self.config.allow_cross_origin() && data.client_data_json.cross_origin.unwrap_or(false)
+        {
+            return Err(WebauthnError::CredentialCrossOrigin);
         }
 
         // Verify that the value of C.tokenBinding.status matches the state of Token Binding for the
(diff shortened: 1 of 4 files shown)


src/proto.rs
```diff
@@ -970,9 +961,17 @@ pub struct CollectedClientData {
     pub challenge: Base64UrlSafeData,
     /// The rp origin as the browser understood it.
     pub origin: String,
+    /// The inverse of the sameOriginWithAncestors argument value that was
+    /// passed into the internal method.
+    #[serde(rename = "crossOrigin", skip_serializing_if = "Option::is_none")]
+    pub cross_origin: Option<bool>,
```
Firstyear, line 967: Do we need to perform any validation around the option and it's usage? We should understand what it does ....

Firstyear (approved):
Looks good, just want to understand what cross_origin does, and if we need to validate anything about it.

src/proto.rs
ericmarkmartin: I think it indicates if the javascript context that created the public key credential had a different origin than the origin that the credential is created for.

We could consider adding a predicate `allow_cross_origin` in the `WebauthnConfig` trait (defaulting to `false` always) that the user can set to indicate whether they want to accept such credentials.

ericmarkmartin:
So I think `cross_origin` just indicates that the credential was created by javascript running on an origin different than that of the RP the credential was made for. I think it might be worth allowing users to disallow this behavior, but interested in your thoughts.

Firstyear:
I think that sounds correct, we should have an allow_cross_origin bool in the config trait. By default we disallow it?

ericmarkmartin:
Yeah I'd think so--I'll push new changes my tomorrow

src/core.rs
```diff
@@ -297,7 +297,10 @@ impl<T> Webauthn<T> {
         }
 
         // Verify that the value of C.origin matches the Relying Party's origin.
-        if data.client_data_json.origin != self.config.get_origin() {
+        if !self.config.allow_cross_origin()
```
Firstyear, line 300: I think this is not correct. https://www.w3.org/TR/webauthn-2/#sctn-verifying-assertion does not actually inspect crossOrigin during the assertion or registration, and we still need to check the origin is our rp origin.

So this makes me question, what is even the point of this boolean if we never inspect it during a registration/assertion? 

@agrinman Thoughts?

ericmarkmartin: I think there's a bit [here](https://www.w3.org/TR/webauthn-2/#clientdatajson-verification) about limited verification if you cannot support a full JSON parser?

ericmarkmartin: > Verifiers may use the following algorithm to verify an encoded CollectedClientData if they cannot support a full JSON parser

Firstyear: Step 4 there still states:

```
A string, origin, that contains the expected origin that issued the request to the user agent.
```

The issue is that your logic as it's set, if you allow cross_origin, then the origin value isn't checked. This is incorrect.

The code should be:

```
if data.client_data_json.origin != self.config.get_origin() {
    // as before
}

if !self.config.allow_cross_origin() && data.client_data_json.cross_origin.unwrap_or(false) {
    // return error, dis allowing cross origin ccd
}
```

ericmarkmartin: That makes sense

Issue #74, opened 2021-06-01
Rename COSE content type to algorithm


madwizard-thomas:
Moved from #73:
> Some other minor things I came across is COSEContentType: this is called COSE algorithm in WebAuthn and COSE specs. Also ECDSA_SHA256/ECDSA_SHA384/ECDSA_SHA512 are called ES256, ES384 and ES512 in the specs.

The algorithm simultaneously defines the message digest algorithm (e.g. SHA256) and the signature algorithm (ECDSA, EDDSA, RSA-PSS, etc.). Implicitly this also restricts things like the type of key and curve used.

Firstyear:
I think this is resolved

PR #75, opened 2021-06-01
Rename COSE content type and ESDSA members to conform to spec

madwizard-thomas:
Implements #74

- [x] cargo fmt has been run


madwizard-thomas:
May be better to rename the _type members to `alg` or `algorithm` as well (`alg` is usually used in the spec)

Issue #73, opened 2021-05-21
Requirements to finalise v0.3


Firstyear:
What would be required to finalise v0.3? Are there any other breaking api changes we want to make?

Firstyear:
cc @ericmarkmartin @agrinman

madwizard-thomas:
Some thoughts:

- What do you think about having separate structures containing the attestation statement data (e.g. `PackedAttestation`, `TpmAttestation` etc.) so the data they contain can be accessed later (for example in the trust validation). There could be a `Verify` trait implemented on all of them with a method `verify(authenticatorData, policy)` instead of the separate `verify_*_attestation` methods?
- `UserHandle` has a maximum of 64 bytes in the spec, that might warrant a specific type for it.

Firstyear:
I think the userHandle comment is good, we should check and limit that I suspect.

I'm not sure what you mean by the attestation statement types? We already return the attested credential data, but I'm assuming you mean we should return the ACD in a more user-friendly format for callers to be able to inspect?

madwizard-thomas:
> I'm not sure what you mean by the attestation statement types? We already return the attested credential data, but I'm assuming you mean we should return the ACD in a more user-friendly format for callers to be able to inspect?

Yes I meant separating the parsing and verification, so the parsed data is available for later. Although that may be of limited use,  trust validation mainly uses the attestation type and path. The main reason I mentioned it is the different verify functions stood out and a Verify trait seemed more idiomatic to me.

Some other minor things I came across is `COSEContentType`: this is called COSE algorithm in WebAuthn and COSE specs. Also ECDSA_SHA256/ECDSA_SHA384/ECDSA_SHA512 are called ES256, ES384 and ES512 in the specs.

Should I create separate issues for these kind of things? 

Btw which WebAuthn spec is the library aiming for? v1 or 2? Or both?

Firstyear:
> > I'm not sure what you mean by the attestation statement types? We already return the attested credential data, but I'm assuming you mean we should return the ACD in a more user-friendly format for callers to be able to inspect?
> 
> Yes I meant separating the parsing and verification, so the parsed data is available for later. Although that may be of limited use, trust validation mainly uses the attestation type and path. The main reason I mentioned it is the different verify functions stood out and a Verify trait seemed more idiomatic to me.

[...]

PR #77, opened 2021-06-07
Verify we have implemented Level 2 of RP

Firstyear:
Relates #73 - this is a review of our registration and authentication code to ensure that we have followed the steps as laid out in level 2 of the webauthn specification.


Firstyear:
Since this is comments-only update, I'm going to merge this :)

PR #78, opened 2021-06-07
Support U2F appid extension

dani-garcia:
Implements support for the appid extension for U2F backwards compatibility.

I've noticed that some types have been recently changed to return &str instead of a String, and I've considered doing the same here, but didn't want to complicate the code adding lifetimes, let me know if you'd prefer it that way and I can change it.

Might also want to change the parameter type from String to something like Into<String>, which would make it easier to use both &str and String types directly.

- [x] cargo fmt has been run


Firstyear:
It would be good to have an associated test case with test vectors to make sure this is working for decoding. Otherwise looks nice and simple :)

Firstyear:
Adding @ericmarkmartin to review as they wrote most of the Extension code originally and may have input :)

ericmarkmartin:
It looks good to me, though agreed about the test. My one (very insignificant question) is whether the rust name for the field should be `app_id`.

dani-garcia:
Regarding the name, I've kept it that way to match the expected value in the spec, I can change it to app_id if it's preferred but then I'd also need to add a `#[serde(rename = "appid")]` to keep it from being sent as `appId` when encoded.

I'm not sure I understand what decoding test you have in mind, the `RequestAuthenticationExtensions` object is passed directly in `generate_challenge_authenticate_options` and output to the client as JSON, so this object is only ever encoded, and never decoded. As far as I know the client doesn't send the value back.

I've cheched if the clients send any input with this extension that might need validation and I've noticed that in `authenticate_credential`, we pass a `PublicKeyCredential` received from the client, and when using the appid extension, my client sends this object:
```json
{
   "id":"uXZjCXdT...",
   "rawId":"uXZjCXdTf...",
   "type":"public-key",
   "extensions":{
      "appid":true
   },
   "response":{
      "authenticatorData":"NWJT...",
      "clientDataJson":"eyJjaGFs...",
      "signature":"MEUCICxk..."
   }
}
```

That extensions object is not being decoded in `PublicKeyCredential` at the moment, but the official docs say (https://www.w3.org/TR/webauthn-2/#sctn-appid-extension):
> Returns the value of output. If true, the AppID was used and thus, when verifying the assertion, the Relying Party MUST expect the rpIdHash to be the hash of the AppID, not the RP ID.

So I guess I can also use that instead of an external flag to decide which RPID to use at validation time, though it's not necessary for my use case right now, so maybe it can be left for the future.

Firstyear:
> Regarding the name, I've kept it that way to match the expected value in the spec, I can change it to app_id if it's preferred but then I'd also need to add a `#[serde(rename = "appid")]` to keep it from being sent as `appId` when encoded.

I think it's fine as appid right now. :) 

> 
> I'm not sure I understand what decoding test you have in mind, the `RequestAuthenticationExtensions` object is passed directly in `generate_challenge_authenticate_options` and output to the client as JSON, so this object is only ever encoded, and never decoded. As far as I know the client doesn't send the value back.
> 
> I've cheched if the clients send any input with this extension that might need validation and I've noticed that in `authenticate_credential`, we pass a `PublicKeyCredential` received from the client, and when using the appid extension, my client sends this object:
> 
> ```json
> {
>    "id":"uXZjCXdT...",
>    "rawId":"uXZjCXdTf...",
>    "type":"public-key",
>    "extensions":{
>       "appid":true
>    },
>    "response":{
>       "authenticatorData":"NWJT...",
>       "clientDataJson":"eyJjaGFs...",
>       "signature":"MEUCICxk..."
>    }
> }
> ```
> 
> That extensions object is not being decoded in `PublicKeyCredential` at the moment, but the official docs say (https://www.w3.org/TR/webauthn-2/#sctn-appid-extension):
> 
> > Returns the value of output. If true, the AppID was used and thus, when verifying the assertion, the Relying Party MUST expect the rpIdHash to be the hash of the AppID, not the RP ID.
> 
> So I guess I can also use that instead of an external flag to decide which RPID to use at validation time, though it's not necessary for my use case right now, so maybe it can be left for the future.

I think if this is the case, we should support it now. But I'm trying to consider how this logic will look in our code. 

My current thinking is that in `generate_challenge_register_options` we check for the appId extension. If it's present, we take the appId from the extension, and store it to `RegistrationState` as a new field appid `Option<String>`. Then in `register_credential` we decompose this, and we pass the Option to `register_credential_internal`. If it's present, we check rpIdHash is the hash of one of the appid OR the rp_id. This allows us to satsify that MAY condition.

dani-garcia:
I've implemented the changes you mention against the authentication functions (`generate_challenge_authenticate` and `AuthenticationState`), as this extension is mostly meant to apply for the compatibility with existing credentials, and not so much the creation of new ones. Let me know how the code looks and if you think there's any changes that I should make.

The tests still pass, but it might be a good idea to create an alternative test to `test_authentication` that uses the appid path, I'll look into generating some test data to do it.

Firstyear:
@dani-garcia This looks really good, great work! I think tests would be good to have here still so do you want me to wait for you to add these before we proceed? 

@ericmarkmartin Can I get you to double check/approve if you're happy too?

src/core.rs
```diff
@@ -604,8 +606,17 @@ impl<T> Webauthn<T> {
...
-        if data.authenticator_data.rp_id_hash != self.rp_id_hash {
-            return Err(WebauthnError::InvalidRPIDHash);
+        // Note that if we have an appid stored in the state and the client indicates it has used the appid extension,
+        // we check the hash against this appid instead of the Relying Party
+        let has_appid_enabled = rsp.extensions.as_ref().map(|e| e.appid).unwrap_or(false);
+        if has_appid_enabled && appid.is_some() {
+            if data.authenticator_data.rp_id_hash != compute_sha256(appid.as_ref().unwrap().as_bytes()) {
+                return Err(WebauthnError::InvalidRPIDHash);
+            }
+        } else {
```
ericmarkmartin, line 616: any reason this can't be an `else if`?

src/core.rs
```diff
@@ -604,8 +606,17 @@ impl<T> Webauthn<T> {
         // Token Binding ID for the connection.
 
         // Verify that the rpIdHash in authData is the SHA-256 hash of the RP ID expected by the Relying Party.
-        if data.authenticator_data.rp_id_hash != self.rp_id_hash {
-            return Err(WebauthnError::InvalidRPIDHash);
+        // Note that if we have an appid stored in the state and the client indicates it has used the appid extension,
+        // we check the hash against this appid instead of the Relying Party
+        let has_appid_enabled = rsp.extensions.as_ref().map(|e| e.appid).unwrap_or(false);
+        if has_appid_enabled && appid.is_some() {
+            if data.authenticator_data.rp_id_hash != compute_sha256(appid.as_ref().unwrap().as_bytes()) {
```
ericmarkmartin, line 613: can we match here? Checking `is_some` and then `unwrap`ing seems like an anti-pattern to me.

ericmarkmartin (changes requested):
Other than a couple of nits (see review) and @Firstyear's comment about tests, looks great!

dani-garcia:
@Firstyear @ericmarkmartin I've added a test using the new functionality with my yubico 5 on firefox, and fixed the couple of nits mentioned.

ericmarkmartin (approved):
Looks fantastic

src/core.rs
```diff
@@ -604,7 +606,14 @@ impl<T> Webauthn<T> {
         // Token Binding ID for the connection.
 
         // Verify that the rpIdHash in authData is the SHA-256 hash of the RP ID expected by the Relying Party.
-        if data.authenticator_data.rp_id_hash != self.rp_id_hash {
+        // Note that if we have an appid stored in the state and the client indicates it has used the appid extension,
+        // we check the hash against this appid instead of the Relying Party
+        let has_appid_enabled = rsp.extensions.as_ref().map(|e| e.appid);
+        if let (Some(true), Some(appid)) = (has_appid_enabled, appid) {
+            if data.authenticator_data.rp_id_hash != compute_sha256(appid.as_bytes()) {
```
Firstyear, line 613: https://www.w3.org/TR/webauthn-2/#sctn-appid-extension
10.2: "When verifying the assertion, expect that the rpIdHash MAY be the hash of the AppID instead of the RP ID."

The key word here is MAY. If we have a mixed u2f + webauth credential, and we send appId then we need to expect *either* RP ID, rather than one or the other. So I think this logic may not actually be correct on reflection. 

I think it should be:


```
let appid_hash = if rsp.extensions.as_ref().map(|e| e.appid).unwrap_or(false) {
    appid.map(|id| compute_sha256(id.as_bytes()))
} else {
    None
};

// Appid Hash is now Some(Vec<u8>) if requested, None if not.
debug_assert!(rsp.extensions.as_ref().map(|e| e.appid).unwrap_or(false) == appid_hash.is_some());

if ! (
    data.authenticator_data.rp_id_hash == self.rp_id_hash ||
    Some(data.authenticator_data.rp_id_hash) == appid_hash
) {
    return Err(WebauthnError::InvalidRPIDHash);
};
```

src/core.rs
```diff
@@ -1290,6 +1304,82 @@ mod tests {
...
+        "#;
+        let rsp_d: PublicKeyCredential = serde_json::from_str(rsp).unwrap();
+
+        // Now verify it!
+        let r = wan.verify_credential_internal(
+            &rsp_d,
+            UserVerificationPolicy::Discouraged,
+            &zero_chal,
+            &cred,
+            &Some(String::from("https://testing.local/app-id.json")),
```
Firstyear, line 1382: Probably also worth a duplication of this verify_credential_internal where Option<Appid> is None to show it fails when you don't have appid with this credential response. IE .is_err().

Firstyear (changes requested):

Firstyear:
Sorry to do this, it all looks really good but on review I noticed a logic fault, and an area to extend the tests a little. But like Eric said, it looks great, thank you so much!

dani-garcia:
Fixed that logic issue and added a couple more tests:

- We provide a valid RPID, and set appid extension to true, which should effectively ignore the stored AppId and use the WebAuthn RP.
- We provide an AppId, but the extension is set to false, which should only check against the RPID and fail.

Firstyear:
Great work @dani-garcia - thank you for your patience as we reviewed this, and for adding all the changes. Really appreciate it!

PR #79, opened 2021-06-14
Try from str base64

Firstyear:
Adds tryfrom for Base64UrlSafeData so we can convert from Strings

- [ - ] documentation has been updated with relevant examples (if relevant)


Firstyear:
I do wonder if we should split base64 out to a different crate, or add it to the main base64 crate....

ericmarkmartin:
Yeah it's seems to gain more features every PR---+1 for this convert implementation though

PR #81, opened 2021-06-30
Extend registration metadata options

Firstyear:
This changes what we parse to policy_verify_trust to be clearer, and moves the ParsedAttestationData out to proto since we may choose to expose this to consumers later. 

A lingering question is what we return from `register_credential`. Currently we return `Result<(Credential, AuthenticatorData<Registration>), WebauthnError>` but outside of the extensions in authenticator data, this is not very useful.

I think we should consider changing this to return ParsedAttestationData instead, and moving the Registration extensions into PAD, which is "nicer" for consumers to then consume.

I think PAD should be left as NOT ser/deser because I think it should be up to consumers to save what they want from the structure in a way that is compatible and auditable for them. 

### DIFF
--- a/src/attestation.rs
+++ b/src/attestation.rs
@@ -9,9 +9,9 @@ use crate::crypto;
 use crate::crypto::compute_sha256;
 use crate::error::WebauthnError;
 use crate::proto::{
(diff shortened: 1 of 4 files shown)


Firstyear:
See #73

Issue #87, opened 2021-10-16
InvalidRPOrigin when subdomain's used


yaleman:
Situation:

Two frontends:
- https://kanidm.example.com
- https://internal.kanidm.example.com

Origin on kanidmd: https://kanidm.example.com

Registering token only works when client's pointed at https://kanidm.example.com

Authenticating against https://kanidm.example.com works OK with webauthn.

Authenticating against https://internal.kanidm.example.com fails:

```
TRACE    tide-request [ 156µs | 100.000% ]
INFO     ┝━ 💬 [request.info]: Request received | http.method: "POST" | path: "/v1/auth"
INFO     ┝━ 💬 [security.info]: Begin auth event | sessionid: Some(00000000-616b-5205-34c5-891ea36cd0a9) | req: AuthRequest { step: Cred(Webauthn(_)) }
INFO     ┝━ 💬 [security.info]: Handler::Webauthn -> Result::Denied - webauthn error | e: InvalidRPOrigin
INFO     ┝━ 💬 [security.info]: Credentials denied | reason: invalid webauthn authentication
INFO     ┝━ 💬 [security.info]: Sending auth result | res: Ok(AuthResult { sessionid: 00000000-616b-5205-34c5-891ea36cd0a9, state: Denied("invalid webauthn authentication"), delay: None })
INFO     ┕━ 💬 [request.info]: --> Response sent | status: 200 - OK
```

As discussed, it seems to relate to: https://github.com/kanidm/webauthn-rs/blob/master/src/core.rs#L316

PR #88, opened 2021-10-17
Allow subdomains of origin

Firstyear:
Implements #87 - in some cases we may wish to allow subdomains of origin to use the same credentials. This allows this to occur with a non-default setting.


yaleman:
Works on my test instance :)

Firstyear:
@ericmarkmartin Can I get a sanity check on the logic here? I disable this subdomain behaviour by default, but would rather someone else double check to make sure I haven't busted it.

Issue #22, opened 2020-02-09
Finish support for x5c attestation in packed


Firstyear:
This was started as part of getting ios+5ci support working, however it is blocked on https://github.com/sfackler/rust-openssl/pull/1095  as we need to access cert extensions and oids to complete the process.

agrinman:
what is missing here for attestations? it seemed like packed format worked for me when i tested it.

Firstyear:
Yes packed works, but there are a number of verification paths I recall that are not yet checked. But as mentioned, this is blocking on being able to read some details on certs that is blocked by rust-openssl :(

madwizard-thomas:
What do you think about using [x509_parser](https://docs.rs/x509-parser/0.9.2/x509_parser/)? If you want to switch to ring/webpki in the future (which I think is doable even with the curve coordinate issues you mentioned), those library have even worse support for reading x509 certificates.

Firstyear:
@ericmarkmartin ^ what do you think about x509 parser here?

ericmarkmartin:
seems solid---I can hack around with it and see how it goes. frustrating that rust openssl hasn't stepped up here

Firstyear:
Yeah :( 

But this also isn't a super important area of the code - webauthn attestation only tells you the manufacturer / trustworthiness of a credential. More important is that we have any cryptographic authentication at all, so the majority of deployments honestly are unlikely to care about attestation. So there's no rush :)

Firstyear:
Complete!

PR #89, opened 2021-11-22
Add x5c extension validation

ericmarkmartin:
Uses x509_parser to validate x5c extensions.

Implements #22.

- [x] cargo fmt has been run


Firstyear:
Looks good so far! Do we have any examples of x5c keys to test this with?

ericmarkmartin:
I'll try to dig one up

Firstyear:
We'll need to rebase this soon, since I'm doing a pretty major rework of the library atm.

ericmarkmartin:
Sounds good.

On an unrelated note, there's also a decision we have to make regarding the Apple certificate extension. The data in [the extension is der-serialized ASN1](https://medium.com/webauthnworks/webauthn-fido2-verifying-apple-anonymous-attestation-5eaff334c849):

```asn1
id-ce-AppleAnonymousAttestation OBJECT IDENTIFIER ::= {iso(1) member-body(2) us(840) apple(113635) appleDataSecurity(100) appleWebAuthnPolicies(8) 2}

AppleAnonymousAttestation ::= SEQUENCE {
    nonce                   [1]     OCTET STRING
}
```

In actuality, this results in the extension value being 38 bytes---6 bytes of asn1 headers followed by 32 bytes that is the actual nonce we're supposed to check. I have a working routine using the `der_parser` crate (which is a dependency of `x509_parser` anyway) to parse the extension value in a principled way, but I'm wondering if we aren't just better off dropping the first 6 bytes.

Firstyear:
If we can do this safely with nom, then I don't see an issue with dropping the bytes ourself :)

ericmarkmartin:
Sorry I think I may have been a bit unclear. der_parser is a crate for parsing these structures with nom. By just dropping the bytes I mean literally `extension.value[6..]`

Firstyear:
I think we should have this in a seperate issue :)

ericmarkmartin:
That's part of Apple anonymous attestation verification...so should we leave all of that for another issue as well? Then in this PR we just add the extension validation function and a test to make sure we can actually parse extensions?

Firstyear:
Sounds good!

webauthn-rs/src/proto.rs
```diff
@@ -1216,7 +1216,7 @@ fn acd_parser(i: &[u8]) -> nom::IResult<&[u8], AttestedCredentialData> {
     Ok((
         i,
         AttestedCredentialData {
-            aaguid: aaguid.to_vec(),
+            aaguid: aaguid.try_into().expect("took 16 characters"),
```
Firstyear, line 1219: I don't think we should expect here, since this is untrusted user data which could cause a crash. We should map_err and ? here instead.

Firstyear (changes requested):

webauthn-rs/src/proto.rs
ericmarkmartin: We parsed the aaguid with `take(16usize)(i)`. Shouldn't this have returned an error already if we are unable to get 16?

webauthn-rs-core/src/attestation.rs
```diff
@@ -17,6 +17,86 @@ use openssl::stack;
 use openssl::x509;
 use openssl::x509::store;
 use openssl::x509::verify;
+use x509_parser::oid_registry::Oid;
+
+/// x509 certificate extensions are validated in the webauthn spec by checking
+/// that the value of the extension is equal to some other value
+pub(crate) trait AttestationX509Extension {
```
Firstyear, line 24: Can you implement this for anonca (look for `1.2.840.113635.100.8.2`) because I'm wondering if this trait will "feel awkward" to use, and that's a good test.

webauthn-rs-core/src/attestation.rs
```diff
@@ -134,11 +214,7 @@ pub(crate) fn verify_packed_attestation(
             // (id-fido-gen-ce-aaguid) verify that the value of this extension matches the aaguid
             // in authenticatorData.
 
-            if let Some(aaguid) = get_fido_gen_ce_aaguid(attestn_cert) {
-                if acd.aaguid != aaguid {
-                    return Err(WebauthnError::AttestationCertificateAAGUIDMismatch);
-                }
-            }
+            validate_extension::<FidoGenCeAaguid>(&attestn_cert, &acd.aaguid)?;
```
Firstyear, line 217: Look for `id-fido-gen-ce-aaguid` in the tpm attestation, because under assert_tpm_attest_req, it's also used there.

Firstyear (changes requested):
i'm not sure how I feel about this as a trait yet, but it looks mostly good. Added some comments :)

webauthn-rs/src/proto.rs
Firstyear: Yes, I think so.

webauthn-rs-core/src/attestation.rs
ericmarkmartin: I had mocked this up at one point. I think the following works (if you add the new `WebauthnError` variants that I made up below)
```rust
impl AttestationX509Extension for AppleAnonymousNonce {
    type Output = [u8; 32];

    // 4. Verify that nonce equals the value of the extension with OID ( 1.2.840.113635.100.8.2 ) in credCert. The nonce here is used to prove that the attestation is live and to protect the integrity of the authenticatorData and the client data.
    const OID: Oid<'static> = der_parser::oid!(1.2.840 .113635 .100 .8 .2);

    fn parse(i: &[u8]) -> der_parser::error::BerResult<&Self::Output> {
        use der_parser::{der::*, error::BerError};
        parse_der_container(|i: &[u8], hdr: Header| {
            if hdr.tag() != Tag::Sequence {
                return Err(nom::Err::Error(BerError::BerTypeError.into()));
            }
            let (i, tagged_nonce) = parse_der_tagged_explicit(1, parse_der_octetstring)(i)?;
            let (class, _tag, nonce) = tagged_nonce.as_tagged()?;
            if class != Class::ContextSpecific {
                return Err(nom::Err::Error(BerError::BerTypeError.into()));
            }
            let nonce = nonce
                .as_slice()?
                .try_into()
                .map_err(|_| der_parser::error::BerError::InvalidLength)?;
            Ok((i, nonce))
        })(i)
    }

    fn if_missing() -> Result<(), WebauthnError> {
        Err(WebauthnError::AttestationStatementMissingExtension)
    }

    fn if_unequal() -> WebauthnError {
        WebauthnError::AttestationCertificateNonceMismatch
    }
}
```

Firstyear: Want to add it to the PR? I think that example has shown me I was wrong to doubt you :)

webauthn-rs-core/src/attestation.rs
```diff
@@ -17,6 +17,119 @@ use openssl::stack;
...
+                        if output == data {
+                            Ok(())
+                        } else {
+                            Err(T::VALIDATION_ERROR)
+                        }
+                    })
+            })
+        })
+        .unwrap_or_else(|| {
+            if T::IS_REQUIRED {
```
Firstyear, line 126: I think so far every extension is "required" so we don't need this?

Firstyear (changes requested):
Last minor change then I think it's ready to merge :) Thank you so much!

Firstyear:
@ericmarkmartin Anything else? Or good for me to merge?

ericmarkmartin:
I want to collect a tpm vector with the cert extension to add tests for that

Firstyear:
The problem there is we can't just "get a tpm vector" without access to the hardware. So for now, I'm gonna merge this, and since it's still in dev, we can fix up any rough edges later :)

PR #100, opened 2022-02-22
20220221 refactor

Firstyear:
This is a major refactor to split proto to it's own crate, allow base64 to be it's own crate, and improve some of our interfaces.  I still want to change how we return Authenticator data to allow it to be serialised, and to add a builder pattern to the webauthn type. 


Firstyear:
@ericmarkmartin If you don't mind having a look at this, that would be great. It's a pretty huge change, but it implements a lot of the needed bits for attestation, as well as better wrappers for extensions. I'm not fully done though, but I think it should be merged before it gets further out of hand.

Firstyear:
And https://webauthn.firstyear.id.au/ is already updated on the new code if you want to test :)

ericmarkmartin:
I know it's a bit late, but it looks great!

Firstyear:
@ericmarkmartin Thanks mate, still much more to do, but it's really getting there. Happy to have any feedback during this dev cycle since we are making a bunch of break changes. I'm going to use git main on a few other projects to try and get some experience with it too, but extensions, attestation and simpler workflows are the goal of this cycle.

PR #102, opened 2022-03-03
Extract credential creation from attestation verification functions

madwizard-thomas:
I have extracted the Credential::new calls from the attestation verification functions because they code is largely duplicated and parameters are passed to it that are otherwise unused by the verification procedures. Also it follows the output from the specs more closely:

> Verification procedure: The procedure for verifying an [attestation statement](https://www.w3.org/TR/webauthn-2/#attestation-statement), which takes the following verification procedure inputs:
> - attStmt: The [attestation statement](https://www.w3.org/TR/webauthn-2/#attestation-statement) structure
> - authenticatorData: The [authenticator data](https://www.w3.org/TR/webauthn-2/#authenticator-data) claimed to have been used for the attestation
> - clientDataHash: The [hash of the serialized client data](https://www.w3.org/TR/webauthn-2/#collectedclientdata-hash-of-the-serialized-client-data)
>
> The procedure returns either:
> 
> - An error indicating that the attestation is invalid, or
> - An implementation-specific value representing the [attestation type](https://www.w3.org/TR/webauthn-2/#attestation-type), and the [trust path](https://www.w3.org/TR/webauthn-2/#attestation-trust-path). This attestation trust path is either empty (in case of [self attestation](https://www.w3.org/TR/webauthn-2/#self-attestation)), or a set of X.509 certificates.

`Result<ParsedAttestationData>` would mirror the return value of this procedure better than the whole Credential object.

Also changed verify_attestation_ca_chain to act on the parsed attestation data rather than the credential since it does not need it.

I've created this PR against your refactoring branch but it's probably easier to wait until it is merged.

### DIFF
--- a/webauthn-rs-core/src/attestation.rs
+++ b/webauthn-rs-core/src/attestation.rs
@@ -52,9 +52,7 @@ pub(crate) fn verify_packed_attestation(
     acd: &AttestedCredentialData,
     att_obj: &AttestationObject<Registration>,
     client_data_hash: &[u8],
-    registration_policy: UserVerificationPolicy,
-    req_extn: &RequestRegistrationExtensions,
-) -> Result<Credential, WebauthnError> {
+) -> Result<ParsedAttestationData, WebauthnError> {
     let att_stmt = &att_obj.att_stmt;
     let auth_data_bytes = &att_obj.auth_data_bytes;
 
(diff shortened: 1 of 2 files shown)


Firstyear:
I don't know how I closed this by accident - I merged #100 and it auto-closed this. Can you re-open if still needed?

PR #103, opened 2022-03-04
Extract credential creation from attestation verification functions (2)

biomedia-thomas:
- [x] cargo fmt has been run

I have extracted the Credential::new calls from the attestation verification functions because they code is largely duplicated and parameters are passed to it that are otherwise unused by the verification procedures. Also it follows the output from the specs more closely:

> Verification procedure: The procedure for verifying an [attestation statement](https://www.w3.org/TR/webauthn-2/#attestation-statement), which takes the following verification procedure inputs:
> - attStmt: The [attestation statement](https://www.w3.org/TR/webauthn-2/#attestation-statement) structure
> - authenticatorData: The [authenticator data](https://www.w3.org/TR/webauthn-2/#authenticator-data) claimed to have been used for the attestation
> - clientDataHash: The [hash of the serialized client data](https://www.w3.org/TR/webauthn-2/#collectedclientdata-hash-of-the-serialized-client-data)
>
> The procedure returns either:
> 
> - An error indicating that the attestation is invalid, or
> - An implementation-specific value representing the [attestation type](https://www.w3.org/TR/webauthn-2/#attestation-type), and the [trust path](https://www.w3.org/TR/webauthn-2/#attestation-trust-path). This attestation trust path is either empty (in case of [self attestation](https://www.w3.org/TR/webauthn-2/#self-attestation)), or a set of X.509 certificates.

`Result<ParsedAttestationData>` would mirror the return value of this procedure better than the whole Credential object.

Also changed verify_attestation_ca_chain to act on the parsed attestation data rather than the credential since it does not need it.

### DIFF
--- a/webauthn-rs-core/src/core.rs
+++ b/webauthn-rs-core/src/core.rs
@@ -21,7 +21,7 @@ use url::Url;
 
 use crate::attestation::{
     verify_apple_anonymous_attestation, verify_attestation_ca_chain, verify_fidou2f_attestation,
-    verify_none_attestation, verify_packed_attestation, verify_tpm_attestation, AttestationFormat,
+    verify_packed_attestation, verify_tpm_attestation, AttestationFormat,
 };
 use crate::constants::{AUTHENTICATOR_TIMEOUT, CHALLENGE_SIZE_BYTES};
 use crate::crypto::compute_sha256;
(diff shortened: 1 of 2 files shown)


webauthn-rs-core/src/core.rs
```diff
@@ -470,56 +470,49 @@ impl WebauthnCore {
...
-                        policy,
-                        ParsedAttestationData::Uncertain,
-                        None,
-                    ))
+                    Ok(ParsedAttestationData::Uncertain)
                 }
             }
         }?;
 
+        let credential: Credential = Credential::new(
```
Firstyear, line 507: Since we have the Credential::new now outside of the attestation code, Ithink this should be "after" the call to verify_attestation_ca_chain.

Firstyear (changes requested):
Only one pretty minor comment, otherwise looks really good! Thank you!

webauthn-rs-core/src/core.rs
madwizard-thomas: Maybe even move it after the algorithm and excluded credentials check, construct it in last `Ok()` at the end?

Firstyear: Yep, that's good by me!

madwizard-thomas: Ah I remember why I put it above verify_attestation_ca_chain, attestation_data would need to implement Copy or Credential should use a reference because it is used after move (use in verify_attestation_ca_chain, move in Credential::new). 
The other checks would also need to get the `credentials.*` properties from somewhere else if it were moved all the way to the result statement.

madwizard-thomas:
Thanks for the quick merge - i see i mixed up my work and personal github accounts in this PR but that doesn't really matter.

Firstyear:
No problem mate, appreciate you helping out! I'm putting in a fair bit of work to this library at the moment, so it's great to see others want to make it better too :)

PR #104, opened 2022-03-06
Fix EC2 coordinates to be variable in length

madwizard-thomas:
- [x] cargo fmt has been run

The EC2 keys currently have a fixed length of 32 bytes but this is incorrect, only ES256 uses 32 byte coordinates. ES384 uses 48 bytes, ES512 uses 66. The bug is probably due to this line in the spec (and code comments): *Let x be the value corresponding to the "-2" key (representing x coordinate) in credentialPublicKey, and confirm its size to be of 32 bytes* but this line is part of the U2F verification, which only uses ES256 keys.

This PR adds a test for ES256/384/512 CBOR parsing and also fixes the problem by using a vec for the coordinates rather than a fixed size array.


Firstyear:
This actually will break ES256. The secret is in:

```
            let mut x_temp = [0; 32];
            x_temp.copy_from_slice(x);
```

In the cbor encoding, if the key components are less than 32 bytes, it will NOT zero pad, and will send 30 or 31 bytes. This has been a bug before, and caused random key enrollment failures. So we MUST have static sized buffers with the copy-from-slice, to allow the zero pad to occur. 

So as a result, I can't accept this, because the issue would *also* occur on es384 and 521. To support those properly we need:

* Test vectors from real devices with those keys (or to write a softtoken with them)
* To have a match statement in the code to check what key size we expect.

So I think this needs to be completely reworked sorry :(

madwizard-thomas:
I don't think that is correct, because:

- The COSE/CBOR spec specifically states ["leading zero octets should be preserved"](https://datatracker.ietf.org/doc/html/rfc8152#section-13.1.1). It does not mention trailing zero octets but both would make little sense since it is a random coordinate on the curve and the chances of having leading or trailing zeroes are pretty much random. Canonical CBOR only mentions encoding integers in the smallest possible form but this is a not an integer but a byte string, which contains an encoding of an integer.
- Your existing code already [checked](https://github.com/kanidm/webauthn-rs/blob/5374d7a9d90beefea5ec2afb20c8fb61427b4847/webauthn-rs-core/src/crypto.rs#L325-330) for a hardcoded length of 32: 
   ```rust
   if x.len() != 32 || y.len() != 32 {
       return Err(WebauthnError::COSEKeyECDSAXYInvalid);
   }
   ```
- copy_from_slice does not pad, it [requires](https://doc.rust-lang.org/std/primitive.slice.html#panics-28) source and length to be the same size

Firstyear:
> I don't think that is correct, because:
> 
> * The COSE/CBOR spec specifically states ["leading zero octets should be preserved"](https://datatracker.ietf.org/doc/html/rfc8152#section-13.1.1). It does not mention trailing zero octets but both would make little sense since it is a random coordinate on the curve and the chances of having leading or trailing zeroes are pretty much random. Canonical CBOR only mentions encoding integers in the smallest possible form but this is a not an integer but a byte string, which contains an encoding of an integer.

Sadly we live in a world where specifications are not always followed. 

> * Your existing code already [checked](https://github.com/kanidm/webauthn-rs/blob/5374d7a9d90beefea5ec2afb20c8fb61427b4847/webauthn-rs-core/src/crypto.rs#L325-330) for a hardcoded length of 32:
>   ```rust
>   if x.len() != 32 || y.len() != 32 {
>       return Err(WebauthnError::COSEKeyECDSAXYInvalid);
>   }
>   ```

That code exists because of past failures. 

> * copy_from_slice does not pad, it [requires](https://doc.rust-lang.org/std/primitive.slice.html#panics-28) source and length to be the same size

But we copy into a pre-zero allocated buffer :) 

Still, my point stands that I'd want this separated up inside an alg match so that we can still make assertions about the size of what coords we received, rather than just "trusting" whatever the client sent us is going to be valid.

Firstyear:
First, lets split this up based on a match around which alg we use, that we can still assert the length bounds I think.

Firstyear:
We also need test vectors for this too, and currently I don't think there are any es384 keys in the wild?

madwizard-thomas:
Sorry I'm confused, do you want to reject incorrect lengths or 'fix' them by adding padding?

I think you may have missed this piece of code?
```rust
let coord_len = curve.coordinate_size();
if x.len() != coord_len || y.len() != coord_len {
```
and

```rust
impl ECDSACurve {
    /// Returns the size in bytes of the coordinate components (x and y) for the specified curve
    pub(crate) fn coordinate_size(&self) -> usize {
        match self {
            Self::SECP256R1 => 32,
            Self::SECP384R1 => 48,
            Self::SECP521R1 => 66,
        }
    }
}
```

madwizard-thomas:
> We also need test vectors for this too, and currently I don't think there are any es384 keys in the wild?
I don't think so, but I included hand built CBOR test vectors for it in the code and the coordinates are valid keys (generated by openssl).

Firstyear:
> Sorry I'm confused, do you want to reject incorrect lengths or 'fix' them by adding padding?
> 
> I think you may have missed this piece of code?
> 
> ```rust
> let coord_len = curve.coordinate_size();
> if x.len() != coord_len || y.len() != coord_len {
> ```
> 
> and
> 
> ```rust
> impl ECDSACurve {
>     /// Returns the size in bytes of the coordinate components (x and y) for the specified curve
>     pub(crate) fn coordinate_size(&self) -> usize {
>         match self {
>             Self::SECP256R1 => 32,
>             Self::SECP384R1 => 48,
>             Self::SECP521R1 => 66,
>         }
>     }
> }
> ```

You're right I did miss that. Let me re-read this tomorrow then with fresh eyes :)

Firstyear (approved):

Firstyear:
Yep, rereading this you are absolutely correct. Sorry about the noise, and thank you for your patience and submission as always :)

PR #115, opened 2022-04-06
114 108 detect rs1 passkeys

### LINKED ISSUE #108: Add flag to allow rejecting RS1 signatures
We should have a flag to allow rejecting RS1 signatures. This will cause *some* classes of TPM's to be rejected, but most of these are older 1.2 devices (windows now requires tpm2.0)

### LINKED ISSUE #114: Allow detecting passkeys with counter == 0
Passkeys can't have a counter synced, so when the counter is 0, it's probably a passkey. We can use this to filter and reject passkeys potentially ....

Firstyear:
Fixes #114  Fixes #108 - this adds a cfg feature allowing rs1 to be disabled. Additionally, an experimental internal only flag is added to reject credentials with a counter of 0 which likely indicates a passkey and a credential that is synced between multiple devices. 

### DIFF
--- a/webauthn-rs-core/Cargo.toml
+++ b/webauthn-rs-core/Cargo.toml
@@ -12,6 +12,11 @@ keywords = ["webauthn", "authentication"]
 categories = ["authentication", "web-programming"]
 license = "MPL-2.0"
 
+[features]
+insecure_rs1 = []
+
+default = ["insecure_rs1"]
+
 [dependencies]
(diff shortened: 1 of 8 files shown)


Firstyear:
I'm not clear why these are failing, I'll have a look but they all worked and passed for me. So I'll re-test once we merge?

Firstyear:
@ericmarkmartin Rebased with all your changes, ready for review now :)

PR #136, opened 2022-06-07
Android-SafetyNet & Android-Key Attestation, plus other small additions

agrinman:
Hello! This PR adds several new features and also looking for feedback.

- Adds support for 2 new attestation types: Android-key, Android-SafetyNet
- Adds the attestation format used to the credential
- Adds new CAs
- Adds option to relax port matching on origin checks
- Exposes `core` to be used for more advanced options to library users
- Exposes cert chain validation method so it can be run by library users with various certs

# Todo:
- [ ] Support customized requirements for Android-key attestation based on library user requirements

# Other
- [x] cargo fmt has been run


webauthn-rs-core/src/interface.rs
```diff
@@ -659,6 +663,67 @@ impl AttestationCa {
...
         }
     }
+
+    /// Android ROOT CA 1
+    pub fn android_root_ca_1() -> Self {
+        AttestationCa {
+            ca: x509::X509::from_pem(ANDROID_ROOT_CA_1).unwrap(),
+            platform_only: false,
+            key_storage: KeyStorageClass::SingleDeviceWrappedKey,
+            strict: true,
```
Firstyear, line 673: These shouldn't be in strict, google has been a hostile upstream and has unresolved issues (IE the pixel straight doesn't work)

webauthn-rs/src/lib.rs
```diff
@@ -84,6 +90,13 @@ pub struct Webauthn {
     algorithms: Vec<COSEAlgorithm>,
 }
 
+impl Webauthn {
+    /// For advanced users allow direct access to the core reg/auth methods
+    pub fn advanced(&self) -> &WebauthnCore {
```
Firstyear, line 95: Maybe should be "danger_core_access" or something?

webauthn-rs-core/src/core.rs
```diff
@@ -3056,9 +3108,8 @@ mod tests {
             &RequestRegistrationExtensions::default(),
             false,
         );
-        debug!("{:?}", result);
-        // Currently UNSUPPORTED as openssl doesn't have eddsa management utils that we need.
-        assert!(result.is_err());
+        dbg!("{:?}", &result);
+        assert!(result.is_ok());
```
Firstyear, line 3112: Why does this change to .is_ok() ? We still don't have eddsa?

webauthn-rs-core/src/core.rs
```diff
@@ -1532,6 +1567,7 @@ mod tests {
             registration_policy: UserVerificationPolicy::Discouraged_DO_NOT_USE,
             extensions: RegisteredExtensions::none(),
             attestation: ParsedAttestationData::None,
+            attestation_format: None,
```
Firstyear, line 1570: We don't need an attestation_format flag, because the parsesAttestationData enum indicates the format that wraps it doesn't it?

webauthn-rs-core/Cargo.toml
```diff
@@ -33,6 +33,7 @@ rand = { version = "0.8" }
 url = { version = "2", features = ["serde"] }
 x509-parser = "0.13.0"
 der-parser = "7.0.0"
+jwt-simple = "0.11.0"
```
Firstyear, line 36: Can't use jwt-simple as it's not an openssl backed crypto library (and it has other issues .... turns out every jwt library in rust is fucked in some way )

We'll probably need to swap to https://github.com/kanidm/compact-jwt which we maintain.

webauthn-rs-core/src/attestation.rs
```diff
@@ -688,7 +859,337 @@ pub(crate) fn verify_apple_anonymous_attestation(
...
+        CtsProfileMatchFailed,
+
+        #[error("Timestamp too old")]
+        Expired,
+
+        #[error("Time error: {0}")]
+        Time(#[from] std::time::SystemTimeError),
+    }
+
+    use jwt_simple::prelude::*;
```
Firstyear, line 1084: ^ as above, will need to swap to compact_jwt due to the crypto backend. We can make any changes to compact_jwt you need though.

webauthn-rs-core/src/attestation.rs
```diff
@@ -688,7 +859,337 @@ pub(crate) fn verify_apple_anonymous_attestation(
...
+
+        Ok((certs, verified_claims.custom))
+    }(response_str)
+    .map_err(|e| {
+        error!("jwt saftey-net error: {:?}", e);
+        WebauthnError::AttestationStatementResponseInvalid
+    })?;
+
+    // 5. If successful, return implementation-specific values representing attestation type Anonymous CA and attestation trust path x5c.
+    Ok(ParsedAttestationData::Basic(x5c))
```
Firstyear, line 1189: This format seems to have a lot of extra metadata, is any of it worth capturing for auditing?

webauthn-rs-core/src/core.rs
Firstyear: Ahhh as I saw above, it looks like these use "Basic" in parsedAttestationData.

Firstyear (changes requested):
@ericmarkmartin May want to look too. Looks awesome, I think the most annoying part is we can't use the jwt library you pull in, because lol reasons.

Firstyear:
I think I need to double check some things after this merges, especially for "re-attesting" in the future, because that's something we might need to do. For example, an RP needs to go through all their saved keys to find ones that have been issued by a now known-compromised CA for example. I want to flesh out that process beore the next release but it shouldn't block this :)

Firstyear:
Oh, and THANK YOU so much! This is a great bit of work, really appreciate it !

webauthn-rs/src/lib.rs
Firstyear: Oh, and what kind of use-cases do you forsee this for? It's a bit of a sharp edeg, I'm curious what your idea is here :)

agrinman: passwordless/securitykey methods dont seem to provide enough granularity for me, so why not open it up to those who know exactly the params they want :)

webauthn-rs-core/src/attestation.rs
agrinman: will need to dig into this more...

webauthn-rs-core/src/attestation.rs
agrinman: Yeah absolutely, I think I want to make ParsedAttestationData a struct to include the format, and an enum of metadata that was extracted. I want to adjust the VerifiedAttestation thing to pull out the data and do the comparison one level up so we have access to the metadata

webauthn-rs-core/src/core.rs
agrinman: Maybe I missed something but I think the JWT in safetynet attestation now works

webauthn-rs-core/src/interface.rs
agrinman: what does the strict here mean?

webauthn-rs-core/src/core.rs
agrinman: The JWT library I used supports eddsa

webauthn-rs-core/Cargo.toml
ericmarkmartin: I have a draft right now to make compact jwt work with safety net. The main thing is we need support for x5c pubkeys.

Firstyear: We could merge this as is-and then rework to have compact jwt?

webauthn-rs-core/src/attestation.rs
Firstyear: Okay, I think that's a good idea, but maybe a subsequent PR ?

webauthn-rs-core/src/core.rs
Firstyear: Is eddsa required for android?

webauthn-rs/src/lib.rs
Firstyear: Yeah, but then you could just use "core" directly, you don't need this wrapper then?

webauthn-rs-core/src/interface.rs
Firstyear: It means the vendor is "high quality", in terms of trust in their CA, responsiveness to issues, crypto quality, hardware quality etc. It's a "project specific" thing, because I've started to find credentials that are "fido certified" that are trash. 

For example, google here can't be strict, because they are a hostile upstream and don't respond to issues. There are defects in the pixels ability to do webauthn at all, and they are "refusing to act on it".

Firstyear: Of course, people are still free to use and trust them, but I want to provide a "strict" list which is the "best of the best" of webauthn impl so that not everyone has to do their own investigation into the quality of vendors.

webauthn-rs-core/src/core.rs
agrinman: Not sure. I need to dig into this more, but this test flipped after I added safety net support. is there a reason it should fail?

demo_site/webauthn-rs-demo/src/actors.rs
```diff
@@ -50,7 +50,7 @@ impl WebauthnActor {
         let rp_name = rp_name.to_string();
         let rp_id = rp_id.to_string();
         let rp_origin = Url::parse(rp_origin).expect("Failed to parse origin");
-        let wan = unsafe { WebauthnCore::new(&rp_name, &rp_id, &rp_origin, None, None) };
```
Firstyear, line 53: This needs to stay, Webauthn *is* unsafe, because it's trivial to fuck-it-up with security consequences.

webauthn-rs-core/src/core.rs
Firstyear: Ahhh sorry I mis-understood, it's not a lack of eddsa here, it's ES256, so that's okay.

webauthn-rs/src/lib.rs
Firstyear: I made it unsafe because using webauthn "without guide rails" is *dangerous*. There are so many security footguns and so many ways to fuck it up. I wanted anyone who "used' that library to *really* have to acknowledge that risk, and really understand what they are doing. Sadly, unsafe is the only scary keyword way to force people to actually check "why is that unsafe?".

Firstyear:
@agrinman I think the todo list is:

* strict = false on the google ca's (I'm still kind-of working out what our static CA definitions should look like for people to make their own attestation lists, probably should be it's own issue). 
* leave unsafe on Webauthn (for now ... that could be it's own issue).
* Open an issue for enum about the attestation data for future processing.
* Open an issue to swap out the jwt lib, blocking the next release.

Otherwise I'd think this is good to go? I can certainly help with the jwt thing. :) 

Thanks again for your help <3

Firstyear (changes requested):

demo_site/webauthn-rs-demo/src/actors.rs
agrinman: I understand the intention here, but the `unsafe` keyword is a Rust protection not application-logic "construction is safe" protection. Since there is no unsafe rust code here, it truly should not be unsafe. I.e. there's no memory unsafe operations here.

agrinman: we can change the name of the `new` constructor, but I think `unsafe` here will cause strange downstream effects. For instance, there are audits that check how much "unsafe" code exist in dependencies.

Firstyear: That's also kind of the motivation, is that this *should* turn up in an audit, so that people question if it's being used correctly. But I can certainly see how from a "pure memory safety" perspective it's not "accurate". That's why  Ithink this should be a seperate issue to discuss (I'm not saying no, but I think seperate issue to focus on that one thing)

agrinman: it's tough because I think library users will want to customize there own use of WebauthnCore. Most libraries allow full customization of this. I think the safer interfaces are nice, but I needed more customization in my use case.

Firstyear: Yeah, I'm wondering what your customisations are that you need? It may not be "unique" to you :)

agrinman:
Ok so newest commit includes the following changes (some we discussed out of band)

- drop `unsafe` keyword for `Webauthn` constructor, rename with something more appropriate :)
- drop strict on CAs that were added
- Credential now stores a `ParsedAttestation` struct which encompasses metadata and the actual old `ParsedAttestationData` (the chain). Metadata support is limited and can easily be exposed later without breaking changes.

Please let me know if anything I missed!

Follow ups:
- [ ] replace jwt dep with compat-jwt (once it's ready)

Firstyear:
[ ] make a better policy about CA review process that we perform (maybe other grading levels)
[ ] Improve the parsed metadata return process?

Firstyear:
Thank you!

PR #154, opened 2022-07-12
Did some housecleaning for Clippy v2

benwis:
Dug around in the code a bit and eliminated most of the clippy errors I could find. Mostly that consisted of removing redundant .clone()s, references that are destructured, and converting some of the intos into From. There are still some left, but 3 of them require rewriting some functions with tons of inputs, 1 of them is an ugly bit based nom function, and 2 of them refer to Intos that I didn't convert because serde is deriving stuff from them.

- [x] cargo fmt has been run

### DIFF
--- a/webauthn-rs-core/src/internals.rs
+++ b/webauthn-rs-core/src/internals.rs
@@ -113,7 +113,7 @@ impl PartialEq<Credential> for Credential {
         self.cred_id == c.cred_id
     }
 }
-
+#[allow(clippy::too_many_arguments)]
 impl Credential {
     pub(crate) fn new(
         acd: &AttestedCredentialData,
@@ -306,7 +306,7 @@ fn acd_parser(i: &[u8]) -> nom::IResult<&[u8], AttestedCredentialData> {
         },
     ))
 }
-
+#[allow(clippy::type_complexity)]
 fn authenticator_data_flags(i: &[u8]) -> nom::IResult<&[u8], (bool, bool, bool, bool, bool, bool)> {
     // Using nom for bit fields is shit, do it by hand.
     let (i, ctrl) = nom::number::complete::u8(i)?;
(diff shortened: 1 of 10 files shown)


webauthn-rs-core/src/internals.rs
```diff
@@ -1050,16 +1050,16 @@ fn tpmtsignature_parser(input: &[u8]) -> nom::IResult<&[u8], TpmtSignature> {
 /// [1]: https://trustedcomputinggroup.org/wp-content/uploads/TCG-TPM-VendorIDRegistry-v1p06-r0p91-pub.pdf,
 #[derive(Debug)]
 pub enum TpmVendor {
-    AMD,
+    Amd,
```
yaleman, line 1053: Changes here are going to have downstream implications... will need to make it clear when bumping the version

yaleman (commented):
Looks great, I do love 📎-calming. I'll let @Firstyear do the approval since he knows this code *far* better, and the change in the enum needs noting somewhere before releasing this into the wild.

webauthn-rs-core/src/internals.rs
Firstyear: This was added this cycle, so that's okay.

Saying that, these are acronyms of vendors, so they should remain as they were capitalised, and clippy should be told to ignore this :)

Firstyear (changes requested):
Looks all good, only need to revert the vendor name casing :)

webauthn-rs-core/src/internals.rs
yaleman: 💯, ❤️

benwis:
I have reverted the acronym changes and added clippy lints for the other troublesome offenders. The repo is now clippy error free!

Firstyear:
Thank you so much!

PR #164, opened 2022-08-08
update axum-sessions to v0.3.0

maxcountryman:
This brings axum-sessions up to the latest version. The primary change
is how sessions are extracted, namely that axum-sessions now provides a
ReadableSession and WritableSession. Otherwise things remain unchanged
and should continue to function as before.

Note that the axum example currently sets secure to true, which means
the example won't work in some browsers (Safari, for instance) unless
it's in fact secured; Tide sets this to false by default.

Fixes #

- [ ] cargo fmt has been run


Firstyear:
Thank you @maxcountryman !

PR #172, opened 2022-08-12
ISO7816 refactor

micolous:
This significantly refactors the NFC related code, adds many references, and works around some broken things.

* Add new `iso7816` module, which provides a slightly higher level
  abstraction around the request and response protocol, so it doesn't have
  APDU constants anymore.

* Add new `atr` module, which parses the answer-to-response bytes.  This is
  needed for detecting if extended Lc/Le is supported, but then Yubikey
  appears to lie about that.

* Add new `tlv` module, which parses Simple-TLV.  This is needed to parse
  the ATR.

* Move CBOR related parsing code out of `nfc::apdu` into its own module
  (`cbor`).  Most of the code is copied as-is, but is renamed for simplicity
  and consistency.  Each of the message types is in its own file.

* Add `GetInfoRequest` struct.

* Rename `ToNfcApdu` to `CBORCommand`, and deliniate bits by transport.

* Move ISO7816 request chunking code into `CBORCommand::to_short_apdus()`.

* Always use `serde` for seralisation, even when NFC is not built.

* Adds a bunch of new tests and documentation.

Follow ups:

* `conformance` example, which detects tokens that don't implement ISO 7816 correctly

Fixes #

- [x] cargo fmt has been run


webauthn-authenticator-rs/src/cbor/make_credential.rs
```diff
@@ -0,0 +1,215 @@
...
+        Integer(6): Map({Text("hmac-secret"): Bool(true)}),
+        // May need to set uv false?
+        Integer(7): Map({Text("rk"): Bool(true)}),
+        // Not needed
+        Integer(8): Bytes([252, 67, 170, 164, 17, 217, 72, 204, 108, 55, 6, 139, 141, 161, 213, 8]),
+        // Not needed
+        Integer(9): Integer(1)}))
+        */
+
+        // why is the a prepended 0x01?
```
Firstyear, line 164: There is a reason for it buried in the ctap spec, and I forgot where ...

Firstyear (approved):

webauthn-authenticator-rs/src/cbor/make_credential.rs
micolous: Yeah, this was something I copied from existing code. The reason is it's the error code :)

micolous:
This is ready to merge now.

Firstyear (approved):

PR #176, opened 2022-08-16
20220816 updates for tpm fido

Firstyear:
Add support to store the TPM version in it's attestation metadata. this can be used in conjuction with the MDS to determine if the device is on an FW version that is trustworthy


Firstyear:
@yaleman It's a bit of random stuff, it's hiding some stuff about attestation ca lists because I'm not willinng to commit to them as a public api yet, but otherwise it's pre-release tidy

yaleman (approved):
🥳

PR #179, opened 2022-08-19
Adding Eq and PartialEq so I can use biscuit claims for the json

chotchki:
Added Eq and PartialEq to types to allow use of PasskeyAuthentication and PasskeyRegistration in biscuit JWE claims. 

For reference: https://docs.rs/biscuit/latest/biscuit/struct.ClaimsSet.html

- [X] cargo fmt has been run
- [X] cargo test has been run and passes
- [N/A] documentation has been updated with relevant examples (if relevant)

### DIFF
--- a/webauthn-rs-core/src/interface.rs
+++ b/webauthn-rs-core/src/interface.rs
@@ -27,7 +27,7 @@ pub type Counter = u32;
 /// The in progress state of a credential registration attempt. You must persist this in a server
 /// side location associated to the active session requesting the registration. This contains the
 /// user unique id which you can use to reference the user requesting the registration.
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
 pub struct RegistrationState {
     pub(crate) policy: UserVerificationPolicy,
     pub(crate) exclude_credentials: Vec<CredentialID>,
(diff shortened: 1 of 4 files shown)


benwis:
Out of curiosity, what does Eq and Partial Eq give you with regard to biscuit JWE claims?

Firstyear:
Looks like they want to stuff the reg and auth state into biscuit tokens rather than storing it server side.

Firstyear:
I'm not really sure how I feel about this PR. Yes, it solves your immediate problem, but I think it seems more like an issue in biscuit, because these aren't really items that actually need or should have to support equality. Can you send some sample code so that I can look at?

Firstyear:
@chotchki Hey there, do you have some sample code I can look at?

Firstyear:
@chotchki I think having a look at this, despite your 5 minute expiration window you still may be open to short lived replay attacks since the client controls the complete state of the authentication flow in this. So it's "not a great idea" to proceed with this design. These state elements *MUST* be stored server side. 

Additionally, you don't actually need to derive PartialEq here to this library. If you implement PartialEq manually rather than deriving it on your various Claims structures, that satisfies the trait bounds for ClaimSet. 

I will document this in the library, so that other people don't fall into the same trap, but sadly it's something we can't easily "prevent" occurring. It may even be worth removing serialise/deserialise on the state types to remove the "illusion" they can be persisted in a token like this. 

So I'm sorry, but I'm going to close this, and I have to advise that you redesign your integration to store the authentication and registration state to be server side stored OR that you manually implement partial eq on your ClaimSet types rather than deriving it.

chotchki:
I appreciate you taking a look! I tend to avoid sessions since they end up using cookies and you then have to implement additional cross site scripting protections but you're motivating me to better understand the attack surface for webauthn.

Firstyear:
No problem at all! Webauthn is certainly "not perfect" and there are ways to still hold it wrong, and the goal of this library is to minimise those. So thanks for reporting the issue :)

Firstyear:
PS: Don't feel bad either, it looks like the project was being naughty in some of it's demo code, so now I have to fix that! Opps!

PR #180, opened 2022-08-24
Reduce ability for people to accidentally allow replay attacks

### LINKED ISSUE #170: Possible stricter security check credential counter
Correct me if I'm wrong, but I was looking at the code and came across this line.

https://github.com/kanidm/webauthn-rs/blob/b53ddfe4770dd34abad3d86512099e4671186c50/webauthn-rs-core/src/core.rs#L1019

Wouldn't it be fair to assume that if the counter of a credential once had a value above 0 (as persisted on the server) it should not be possible for a subsequent request to have the counter be 0 again.

If that assumption is correct maybe this check can be a bit stricter by removing the `|| cred.counter > 0` check?

Firstyear:
Fixes #170 - Help to document the risks of serialising the reg/auth states with client side held cookies. 


yaleman (approved):
I do like me some security warnings! 🥳

PR #189, opened 2022-09-07
Improve remaining docs

### LINKED ISSUE #187: Rethinking the feature `danger-allow-state-serialisation`
While updating webauthn-rs from 0.3 -> 0.4 I noticed the same thing as #184 where the removal of `serde(Serialize, Deserialize)` from states meant they could no longer be stored in the database, cache, or anything along those lines. As far as I can tell the only place you can store it _without_ enabling the flag in some sort of `HashMap`. *Even both examples needs to enable it in order to actually provide full functionality.*

I understand the thought behind making it a feature, and making it sound scary with "danger-" in the name of the feature. So users have to be aware of the client side warnings before enabling a potential buggy case.

However, if even the most basic examples have to enable this feature the chance someone jumps into a codebase by someone who has it enabled because of necessity becomes very high. Not to mention it adds quite a bit of friction to users just jumping in. Since they can't follow the big bold warning labels that say "you must store this on the server side" without finding a complete separate feature (for example the docs: https://docs.rs/webauthn-rs/0.4.5/webauthn_rs/struct.Webauthn.html#method.start_securitykey_registration which mention the warning but don't mention the serialization features).

I'm hoping to open some discussion around how we can lower the friction, while hopefully keeping similar amount of warnings. Perhaps some ideas might be like:

 - adding in features for specific databases (such as implementing `#[sqlx::Type]` if you enable a `sqlx` feature, `#[SqlType]` for a `diesel` features, etc.
 - enabling the feature by default, but changing method names to be more clear something is not safe for the client side. (or just trusting the documentation).

These are just basic ideas, and I'm sure there are much smarter ones. I just really want to open a discussion somewhere.

Firstyear:
Fixes #187 significantly improves the docs related to seralisation. 

- [ ] cargo fmt has been run

### DIFF
--- a/fido-mds/src/mds.rs
+++ b/fido-mds/src/mds.rs
@@ -1,5 +1,5 @@
 //! An implementation of the types for the fido metadata service as defined by
-//! https://fidoalliance.org/specs/mds/fido-metadata-service-v3.0-ps-20210518.html
+//! <https://fidoalliance.org/specs/mds/fido-metadata-service-v3.0-ps-20210518.html>
 //!
 //! This allows parsing the fido metadata blob and consuming it's content. See `FidoMds`
 //! for more.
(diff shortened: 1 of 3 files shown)


yaleman (approved):
woot

PR #192, opened 2022-09-15
adding codeowners

yaleman:
This is part of the protected branch magic


Firstyear:
I think we need @agrinman as well as they are also a "write" role on the repo and trusted in the library.

yaleman:
Done!

Firstyear (approved):

PR #198, opened 2022-09-23
Use matrixes for CI

micolous:
Use [matrixes](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs) to select the Rust version used by CI, to make it easier to test different features in the future.

This doesn't touch any of the Rust code.


yaleman (approved):

yaleman:
Thanks @micolous

PR #199, opened 2022-09-23
RFC: authenticator library design proposal

micolous:
As previously discussed offline with @Firstyear, this documents the current state of the world for `webauthn-authenticator-rs`, and outlines some future work.


Firstyear (approved):
I think this is an excellent design and goal :)

PR #201, opened 2022-09-23
Serde wasm bindgen error

Firstyear:
Appears to fail when mediation is not sent from the server. I suspect this is because we "skip" the value, and the abscence doesn't map to a variant. Instead we can always send the value. Perhaps a field alias or variant alias would also resolve?

Fixes #

- [ ] cargo fmt has been run


yaleman (dismissed):
I ated the purple berries 😆

yaleman:
(The CI failures are fixed in https://github.com/kanidm/webauthn-rs/pull/200)

yaleman (approved):

PR #204, opened 2022-09-25
Add support for Windows 10 WebAuthn API

micolous:
This adds support for Windows 10's WebAuthn API. This is effectively mandatory on Windows 10 Build 1903 and later, as it blocks direct access to FIDO authenticators for non-Administrator users.

https://user-images.githubusercontent.com/246847/194449801-4e6b89e5-0465-4dea-8dc9-bf2508373147.mp4

This also adds a new `authenticate` example, pulled from `SoftToken`'s unit tests.

Tasks remaining:

- [x] Finish populating `WEBAUTHN_AUTHENTICATOR_MAKE_CREDENTIAL_OPTIONS`
- [x] Implement `perform_auth`
- [x] Come up with a reliable way to get/pass the `HWND` for the application window
- [x] Find a way to pass in application info
- [x] There's probably lifetime issues here, lots of raw pointers... Microsoft's `windows` crate has a very `unsafe` API
- [x] Make new test program for making test credentials (rather than stomping nfc_token_info...
- [x] Remove `println!` debugging
- [x] Set up CI on Windows (#212)
- [x] Split all the non-Windows changes out of this PR (#211 #210)

Fixes #

- [x] cargo fmt has been run

### DIFF
--- a/webauthn-authenticator-rs/src/win10/gui.rs
+++ b/webauthn-authenticator-rs/src/win10/gui.rs
@@ -1,0 +1,271 @@
+//! GUI code for Windows WebAuthn API.
+//!
+//! Windows WebAuthn API needs a [HWND] parameter to know where to attach its
+//! UI to, but the `webauthn-authenticator-rs` API doesn't know anything about
+//! [HWND]s.
+//!
+//! Also, when running in a console application, getting a _functional_ [HWND]
+//! is a broken mess, because [GetConsoleWindow] only works properly for
+//! "classical" `cmd.exe` prompts: Windows Terminal
+//! [has focus issues, even though the issue is "closed"][terminal], and VS
+//! Code [also has z-order issues][vscode]. Both make for an awful experience,
+//! and [the normal work-arounds][hack] don't work when a console app can't
+//! set the window title verbatim (such as in VS Code).
+//!
+//! There is some prior art in [windows-fido-bridge], but this ended up with
+//! a distinct design that doesn't hard-code work-arounds for single apps.
+//!
+//! **Warning:** This hasn't yet been tested in a Windows GUI application,
+//! and is subject to change based on user feedback.
+//!
+//! [Window] creates a 1x1 pixel window for Windows WebAuthn API to use:
+//!
+//! * If there is a [current foreground window][GetForegroundWindow],
+//!   our window is a child tool window, and is centred over it. This gives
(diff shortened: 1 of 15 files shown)


Firstyear:
What else would you want before we merge this?

micolous:
> What else would you want before we merge this?

Once #212 is in, I'll need to rebase this and it should be good to go.

webauthn-authenticator-rs/src/win10/gui.rs
```diff
@@ -0,0 +1,271 @@
...
+        match hwnd {
+            Ok(HWND(0)) | Err(_) => Err(WebauthnCError::Internal),
+            Ok(hwnd) => Ok(Self { hwnd }),
+        }
+    }
+}
+
+impl Drop for Window {
+    fn drop(&mut self) {
+        // trace!("dropping window");
```
Firstyear, line 235: Not that you need to change it, but a trickk in tracing is that you can "compile out" trace level calls so you don't need to comment them out in all cases :)

Firstyear (approved):

webauthn-authenticator-rs/src/win10/gui.rs
```diff
@@ -0,0 +1,271 @@
...
+//! and [the normal work-arounds][hack] don't work when a console app can't
+//! set the window title verbatim (such as in VS Code).
+//!
+//! There is some prior art in [windows-fido-bridge], but this ended up with
+//! a distinct design that doesn't hard-code work-arounds for single apps.
+//!
+//! **Warning:** This hasn't yet been tested in a Windows GUI application,
+//! and is subject to change based on user feedback.
+//!
+//! [Window] creates a 1x1 pixel window for Windows WebAuthn API to use:
```
yaleman, line 21: I wonder if there's any security tools that flag a 1x1 window being popped like that 🤔 time will tell, I guess 😃

webauthn-authenticator-rs/src/win10/gui.rs
micolous: ACK, I ended up commenting it out because it was noisy, but it was useful when writing it to make sure everything lived long enough, and the UI event loop was still running.

webauthn-authenticator-rs/src/win10/gui.rs
micolous: 🤷

PR #206, opened 2022-10-05
Fix type for AuthenticatorExtensionsClientOutputs::cred_blob, add related RegistrationExtensionsClientOutputs::cred_protect

micolous:
Per [FIDO v2.1 s12.2 (Credential Blob)](https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-errata-20220621.html#sctn-credBlob-extension), the `authenticatorGetAssertion` authenticator extension output is a Byte String, not a `bool`.

This also adds the related [credProtect extension output](https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-errata-20220621.html#sctn-credProtect-extension).

I don't know how to test this properly, I tried putting in a CredBlob into the compatibility tester, but either some plumbing is missing or browsers don't support this properly :(

Fixes #

- [x] cargo fmt has been run

### DIFF
--- a/webauthn-rs-proto/src/extensions.rs
+++ b/webauthn-rs-proto/src/extensions.rs
@@ -116,7 +116,7 @@ pub struct CredBlobGet(pub bool);
 /// Extension option inputs for PublicKeyCredentialRequestOptions
 ///
 /// Implements \[AuthenticatorExtensionsClientInputs\] from the spec
-#[derive(Debug, Serialize, Clone, Deserialize)]
+#[derive(Debug, Serialize, Clone, Deserialize, Default)]
 #[serde(rename_all = "camelCase")]
 pub struct RequestAuthenticationExtensions {
     /// The `credBlob` extension options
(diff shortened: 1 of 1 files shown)


Firstyear:
Cred blob is being removed, largeBlob is where it's at. 

You also need to be careful here: some extensions are processed differently by the browser compared to the device. I seem to recall credBlob as a bool *is* correct for browsers, but I need to check.

Anyway, a perfect example of this is:
https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-20210615.html#sctn-hmac-secret-extension

Notice the difference between the *client* extension and the *authenticator* extension. Client here actually means "browser" and authenticator means ... the authenticator. 

The webauthn-rs-proto definitions are intended to be correct for the server to browsers/client/user-agent/portal/interdimensional-gateway, rather than for fido. So I think if we were to do this "properly" we need webauthn-authenticator-rs to know how to map from the client versions to the authenticator versions.

Firstyear:
Anyway, IMO we should just yeet credBlob.

micolous:
The way I'm reading the spec is that `credBlob` is a way for an RP to store _at least_ 32 bytes of arbitrary data on a key for a credential.  Then you end up with these types:

* registration:
  * request: byte blob: data to store
  * response: `bool`: whether the blob was stored or not
* attestation:
  * request: `bool`: always `true` to request the blob
  * response: byte buffer: "empty" if there is no `credBlob` for the credential, or the stored data

Unfortunately it looks like none of my test keys have this extension, so yeah, may be better off just removing this feature.

PR #211, opened 2022-10-07
Add common fields from win10 branch

micolous:
* Add `RegistrationExtensionsClientOutputs` fields `hmac_secret`, `cred_protect`, `min_pin_length`
* Add `AuthenticatorTransport::Test`
* Note missing icon for `User` and `RelyingParty`
* Fix `User::name` docstring

Fixes #

- [x] cargo fmt has been run

### DIFF
--- a/webauthn-rs-proto/src/options.rs
+++ b/webauthn-rs-proto/src/options.rs
@@ -74,9 +74,11 @@ pub struct User {
     /// must NOT contain personally identifying information, as this value can NEVER
     /// be changed. If in doubt, use a UUID.
     pub id: Base64UrlSafeData,
-    /// A detailed name for the account, such as an email address.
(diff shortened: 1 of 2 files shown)


webauthn-rs-proto/src/options.rs
```diff
@@ -73,10 +74,11 @@ pub struct User {
     /// must NOT contain personally identifying information, as this value can NEVER
     /// be changed. If in doubt, use a UUID.
     pub id: Base64UrlSafeData,
-    /// The users preferred name for display.
+    /// A detailed name for the account, such as an email address.
```
Firstyear, line 77: Could be worth commenting that these values can change and should never be used as a primary key.

Firstyear (approved):

Issue #138, opened 2022-06-10
Improve policy and process for CA and device validation


Firstyear:
We need a better policy for how we grade CA's and their devices for the "strict" level (and we may need level beside that).

PR #217, opened 2022-10-25
Allow checking of aaguids in attestation code.

Firstyear:
Improves #138 - This allows verification of the attested AAGUID and it's association to a specific CA that provided it. This maps cleanly to the fido MDS where a CA can sign multiple AAGUIDS, but an AAGUID must only be signed by a single CA. This is working towards the ability to do strict attestation policy. 

Next step will be to allow converting the fido-mds structures into the Attestation CA. I think I will also improve the Attestation CA and Attestation CA List code into proper builder patterns to assist with this. 

### DIFF
--- a/base64urlsafedata/src/lib.rs
+++ b/base64urlsafedata/src/lib.rs
@@ -16,8 +16,10 @@
 
 use serde::de::{Error, SeqAccess, Unexpected, Visitor};
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use std::borrow::Borrow;
 use std::convert::TryFrom;
 use std::fmt;
+use std::hash::Hash;
 
 static ALLOWED_DECODING_FORMATS: &[base64::Config] = &[
     base64::URL_SAFE_NO_PAD,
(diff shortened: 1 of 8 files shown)


yaleman (approved):

aseigler:
> This maps cleanly to the fido MDS where a CA can sign multiple AAGUIDS, but an AAGUID must only be signed by a single CA.

From https://fidoalliance.org/specs/mds/fido-metadata-statement-v3.0-ps-20210518.html#dom-metadatastatement-attestationrootcertificates:

attestationRootCertificates, of type DOMString[]
List of attestation trust anchors for the batch chain in the authenticator attestation. Each element of this array represents a PKIX RFC5280 X.509 certificate that is a valid trust anchor for this authenticator model. Multiple certificates might be used for different batches of the same model. The array does not represent a certificate chain, but only the trust anchor of that chain. A trust anchor can be a root certificate, an intermediate CA certificate or even the attestation certificate itself.

So a CA can sign for multiple AAGUIDs, and an AAGUID might be signed by any of the CAs in the `attestationRootCertificates` or any subordinate of any of those, or itself.

Firstyear:
@aseigler Re-checking the code, because a ca has an associated map of aaguids, you can have ca X -> aaguid A and ca Y -> aaguid A, but if you also trust ca Z -> aaguid B, aaguid A may not be associated with this CA.

Does that help clarify?

PR #232, opened 2022-12-04
Add basic support for caBLE authenticators / initiators

micolous:
This implements a subset of caBLE v2.1 based on Chromium's implementation, and can act as both an initiator (acting as a transport for `CtapAuthenticator`) and an authenticator (proxying requests to `AuthenticatorBackendHashedClientData`, handling PIN/UV auth), and is the MVP for #259.

This handles all the Websockets and Noise-like tunnelling needed to make things work:

* If you're using this as an initiator (to authenticate with a mobile device), this generates a QR code, and works with platform Bluetooth APIs (bluez / CoreBluetooth / WinRT) to listen for the service data advertisement.
* If you want to use this as an authenticator (to pretend to be a mobile device) you'll need to provide a way to send service data advertisements (as macOS and Windows don't allow this); there's an example in `cable_tunnel` which uses a serial-connected Bluetooth HCI adapter. It's also best to do this on a separate device from the initiator, because it'll conflict with the initiator's local device access unless you're using `SoftToken`.

This works for me as an initiator with Android and iOS 16 authenticators, and as an authenticator with Chrome and Safari initiators (using Google's tunnel server).

Deferred work listed in #259.

- [x] cargo fmt has been run
- [x] cargo test has been run and passes
- [x] documentation has been updated with relevant examples (if relevant)

<details>
  <summary><strong>In-development screenshots</strong></summary>
  
  ### MakeCredential flow on iOS 16:
  
  MakeCredential QR scanning on iOS
  
  MakeCredential TouchID prompt
  
  ### GetAssertion flow on iOS 16:
  
  GetAssertion QR scanning on iOS
  
  GetAssertion TouchID prompt

</details>

PR #251, opened 2023-01-07
Common API changes needed to (later) add caBLE support.

micolous:
This adds functionality, changes APIs and refactors `webauthn-authenticator-rs` in ways that's necessary (or useful) for caBLE support (#232), without _actually adding_ support for caBLE itself.

* `AuthenticatorBackendHashedClientData`: new trait for `AuthenticatorBackends` which accepts `client_data_hash` directly. This de-duplicates the `client_data_json` handling, and allows the library to proxy requests from a caBLE initiator.
* `perform_register_with_request`, `perform_auth_with_request`: Accepts `MakeCredentialRequest` and `GetAssertionRequest` commands directly, mapping them onto `AuthenticatorBackendHashedClientData`.  This allows it to act as a caBLE authenticator, while _also_ handling PIN/UV auth internally when proxying to physical authenticators (as browsers don't attempt it!)
* crypto: new module for common cryptographic operations, shared between caBLE and CTAP2 implementation.
* CTAP2:
  * implement `AuthenticatorBackendHashedClientData` instead of `AuthenticatorBackend`
  * adds serialisation for `Map<u32, String>` to `GetAssertionResponse` and `MakeCredentialResponse` and from `GetAssertionRequest` and `MakeCredentialRequest` for `perform_*_with_request`.
  * allow for both `Map<String, String>` and `Map<u32, String>` representations of `CBORRequest` and `CBORResponse` for `perform_*_with_request`.
  * add `GetInfoResponse` serialisation
  * make `to_short_apdus` and `to_long_apdu` stand-alone methods
  * `Token` is now mutable, and `Token::close` is now async
  * use short APDUs for selecting FIDO2 applet over NFC, because Feitian tokens don't support extended APDUs (in violation of the FIDO specification!)
* SoftPasskey:
  * implement `AuthenticatorBackendHashedClientData` instead of `AuthenticatorBackend`
* SoftToken: 
  * implement `AuthenticatorBackendHashedClientData` instead of `AuthenticatorBackend`
  * add support for persisting the SoftToken (as `SoftTokenFile`)
  * add `examples/softtoken` for creating a persisted `SoftToken`
* ui: Add methods for caBLE (displaying QR codes and reporting status)
* `examples/authenticate`:
  * refactor to use command line arguments, rather than interactive text UI
  * re-open the connection to the key after every operation (needed for Feitian NFC keys, and also caBLE)
* clean up some dead/commented code that was from testing earlier features
* fix some documentation formatting errors

Fixes #

### DIFF
--- a/webauthn-authenticator-rs/examples/authenticate/main.rs
+++ b/webauthn-authenticator-rs/examples/authenticate/main.rs
@@ -1,17 +1,28 @@
 #[macro_use]
 extern crate tracing;
 
+use std::fs::OpenOptions;
 use std::io::{stdin, stdout, Write};
 
+use clap::{Args, Parser, Subcommand};
 use futures::executor::block_on;
 use webauthn_authenticator_rs::ctap2::CtapAuthenticator;
 use webauthn_authenticator_rs::prelude::Url;
-use webauthn_authenticator_rs::softtoken::SoftToken;
+use webauthn_authenticator_rs::softtoken::{SoftToken, SoftTokenFile};
 use webauthn_authenticator_rs::transport::*;
+use webauthn_authenticator_rs::types::CableRequestType;
 use webauthn_authenticator_rs::ui::{Cli, UiCallback};
 use webauthn_authenticator_rs::AuthenticatorBackend;
 use webauthn_rs_core::proto::RequestAuthenticationExtensions;
 use webauthn_rs_core::WebauthnCore as Webauthn;
+
+#[derive(Debug, clap::Parser)]
+#[clap(about = "Register and authenticate test")]
+pub struct CliParser {
+    /// Provider to use.
+    #[clap(subcommand)]
+    provider: Provider,
+}
 
 fn select_transport<'a, U: UiCallback>(ui: &'a U) -> impl AuthenticatorBackend + 'a {
     let mut reader = AnyTransport::new().unwrap();
@@ -22,9 +33,8 @@ fn select_transport<'a, U: UiCallback>(ui: &'a U) -> impl AuthenticatorBackend +
             while let Some(card) = tokens.pop() {
                 let auth = block_on(CtapAuthenticator::new(card, ui));
 
-                match auth {
-                    Some(auth) => return auth,
-                    None => (),
+                if let Some(auth) = auth {
+                    return auth;
                 }
             }
         }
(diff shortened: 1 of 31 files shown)


micolous:
This should be good to go now.

webauthn-authenticator-rs/examples/authenticate/main.rs
```diff
@@ -118,8 +145,17 @@ fn main() {
     let cred = wan.register_credential(&r, &reg_state, None).unwrap();
 
     trace!(?cred);
+    drop(u);
+    let mut buf = String::new();
+    println!("WARNING: Some NFC keys need to be power-cycled before you can authenticate.");
```
Firstyear, line 150: Doesn't have to be in this PR, maybe we should check the AAGUID of the device get_info and only issue this warning if it's a known problem device?

webauthn-authenticator-rs/examples/softtoken/main.rs
```diff
@@ -0,0 +1,43 @@
+extern crate tracing;
```
Firstyear, line 1: Some of these commands are actually useful and we should move them to authenticator-cli instead soon.

Firstyear (approved):

webauthn-authenticator-rs/examples/authenticate/main.rs
micolous: The specific case this is handling is someone leaving an NFC key placed on the reader.

The Windows WebAuthn API enforces this (on a software level) for all requests -- once the CTAP backend here gets some more polish I'd like to have it work the same way.

The Feitian keys are the only NFC ones I've seen enforce this in firmware.  I haven't actually figured out whether this is another bug in Feitian's firmware, because right now for NFC keys I can test, it's Feitian vs. Yubikey, and I'll need to read the CTAP spec again to figure out if they got ideas from there.  Seeing what a third vendor does would help.

The Feitian key doesn't seem to care if you repeatedly authenticate without moving the key. It only needs a power-cycle after registration... which makes me think that it may not be intentional.

One of the things I'd like to tie into some MDS work is actually document some of these sorts of issues; because from what I can tell the current FIDO stuff relies a lot on libraries shipping work-arounds.

webauthn-authenticator-rs/examples/softtoken/main.rs
micolous: Agreed

PR #254, opened 2023-01-08
Add actix_web tutorial

myOmikron:
I am working on a tutorial for `actix-web`, any notes or requirements for this are welcome.


Firstyear:
@myOmikron Sounds good! Be sure to also add yourself to CONTRIBUTORS.md, and we are happy to review as you proceed.

myOmikron:
Quick question:
Should I use also include the database part or, like in other tutorials, leave it to the user?

tutorial/server/actix_web/src/startup.rs
```diff
@@ -0,0 +1,47 @@
...
+    let rp_id = "localhost";
+    // Url containing the effective domain name
+    // MUST include the port number!
+    let rp_origin = Url::parse("http://localhost:8080").expect("Invalid URL");
+    let builder = WebauthnBuilder::new(rp_id, &rp_origin).expect("Invalid configuration");
+
+    // Now, with the builder you can define other options.
+    // Set a "nice" relying party name. Has no security properties and
+    // may be changed in the future.
+    let builder = builder.rp_name("Axum Webauthn-rs");
```
Firstyear, line 34: Minor detail, wrong RP name

tutorial/server/actix_web/src/main.rs
```diff
@@ -0,0 +1,56 @@
...
+
+    let (webauthn, webauthn_users) = startup();
+
+    // Build the webserver and run it
+    info!("Start listening on: http://0.0.0.0:8080");
+    HttpServer::new(move || {
+        App::new()
+            .wrap(Logger::default())
+            .wrap(
+                SessionMiddleware::builder(CookieSessionStore::default(), key.clone())
```
Firstyear, line 37: I think the cookie session store here stores the full state in the cookie itself, rather that being a key to a memory map. this means in this example replay attacks are possible, which is not great in an example. 

So I think you may need either add a server-side memory session store in actix_session, or you need to add a hashmap or similar to the app state, and then store a key-id or similar in the cookie instead. Thoughts?

Firstyear (changes requested):
Looks really good excepting the two comments, one of which is sadly a security issue we should fix before merging.

tutorial/server/actix_web/src/startup.rs
myOmikron: Resolved by 1d76fbd3c62ea02ecfa56e15d9a7c5572ec82880

tutorial/server/actix_web/src/main.rs
myOmikron: Implemented in d72ab9702768ce14ba8b3d0d10422986ae92c3d0

Firstyear (approved):
Looks really good - you should consider doing a PR to actix sessions for the memory session type too :)

PR #255, opened 2023-01-08
Support slices for unique user IDs

jun-sheaf:
This PR adds a UUID feature to the crate. This allows users to opt-out of UUIDs.

### DIFF
--- a/webauthn-rs/Cargo.toml
+++ b/webauthn-rs/Cargo.toml
@@ -14,11 +14,13 @@ license = "MPL-2.0"
 features = ["danger-allow-state-serialisation", "danger-user-presence-only-security-keys", "danger-credential-internals"]
 
 [features]
+default = ["uuid"]
 resident-key-support = []
 preview-features = []
 danger-allow-state-serialisation = []
 danger-credential-internals = []
 danger-user-presence-only-security-keys = []
+uuid = ["dep:uuid"]
 
 [dependencies]
 base64urlsafedata = { version = "0.1", path = "../base64urlsafedata" }
@@ -26,4 +28,4 @@ webauthn-rs-core = { version = "0.5.0-dev", path = "../webauthn-rs-core" }
 url = { version = "2", features = ["serde"] }
 tracing = "0.1"
 serde = { version = "1", features = ["derive"] }
-uuid = { version = "1", features = ["v4", "serde"] }
+uuid = { version = "1", features = ["v4", "serde"], optional = true }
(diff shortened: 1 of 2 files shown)


Firstyear:
The use of uuid's here was a very specific choice - to force consumers of this library to "do the right thing" and use unique, static identifiers in a location that is *immutable* on created credentials. 

The issue is that by allowing a slice, we then allow people to be lazy and just put in "usernames" or "emails". What happens when a person changes their name? Changes email? Suddenly we break their authenticator credentials.

UUID's are the correct option here, to force RP's and consumers of this library to use proper and correct unique identifiers in a location where they are immutable relative to the authenticator credential created.

So I'm sorry, we won't accept any change to the library that would allow user experience issues like this.

Firstyear:
PS if you really want to use slices, the underlying webauthn-rs-core library will allow it, but it has many sharp-edges and you need to be more careful and aware as an RP

jun-sheaf:
I am aware of the reasons as I've documented them in the description. It's understood UUID enforces certain characteristics required by the spec, but it does not enforce all of them. For example, UUID v5 does not enforce user handles to be unique across different accounts.

If this library really wanted to enforce UUIDs in the way that it wanted, it should do so by not allowing users to put their own UUIDs and just generate a v4 UUID when the method is called and returning it.

In any case, if you want to remain a versatile library, you must allow the use case; it's not part of the spec to enforce UUIDs.

Firstyear:
> 
> In any case, if you want to remain a versatile, unopinionated library, you should allow the use case; it's not part of the spec to enforce UUIDs.

See, we don't want to be a versatile and unopinionated library. The whole point of the webauthn-rs layer is that it is focused on usecases and has opinions that make it impossible to hold incorrectly since the webauthn standard is very prone to misuse that can lead to user security bypasses, user frustration,  cultural and social insensitiveness, and more.

> Regarding using "core" directly, if I had to end up using "core" directly, I'd rather just build everything from scratch since you are enforcing opinionated requirements here; I'd be paranoid of where others devils could be hiding.

Webauthn-rs is a very thin layer over the core library anyway. The core library is more a reflection of the "specification" with the primary exception's of spec-opinion's being documented clearly in the source code for example https://github.com/kanidm/webauthn-rs/blob/master/webauthn-rs-core/src/core.rs#L627 

And in most cases, I have been working with the webauthn wg to have these "out of spec" opinions (as you call them) become part of the spec now (for example, https://github.com/kanidm/webauthn-rs/blob/master/webauthn-rs-core/src/core.rs#L744 ). Sometimes those opinions are there to protect people and enforce requirements the spec hasn't evolved to encompass yet. 

Regardless, feel free to either use core since it fits your use case and audit it for those "opinions" you are so worried about.

PR #256, opened 2023-01-11
Fail builds when running on old/broken versions of OpenSSL

micolous:
This checks that OpenSSL has this patch: https://github.com/openssl/openssl/pull/12826

Old versions of OpenSSL report a `malloc failure` part way through a caBLE handshake.

The existing `hkdf_chromium` test reproduces the issue reliably, so just run that test in `build.rs`.

Fixes #

### DIFF
--- a/webauthn-authenticator-rs/Cargo.toml
+++ b/webauthn-authenticator-rs/Cargo.toml
@@ -58,3 +58,6 @@ tracing-subscriber = { version = "0.3", features = ["env-filter", "std", "fmt"] 
 clap = { version = "^3.2", features = ["derive", "env"] }
 tokio = { version = "1.22.0", features = ["sync", "test-util", "macros", "rt-multi-thread", "time"] }
 tempfile = { version = "3.3.0" }
+
+[build-dependencies]
+openssl = "0.10"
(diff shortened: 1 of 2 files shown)


webauthn-authenticator-rs/Cargo.toml
```diff
@@ -58,3 +58,6 @@ tracing-subscriber = { version = "0.3", features = ["env-filter", "std", "fmt"]
 clap = { version = "^3.2", features = ["derive", "env"] }
 tokio = { version = "1.22.0", features = ["sync", "test-util", "macros", "rt-multi-thread", "time"] }
 tempfile = { version = "3.3.0" }
+
+[build-dependencies]
+openssl = "0.10"
```
yaleman, line 63: Can this bump to the version we just patched over in the other repo, for consistency/avoiding recompilation/other issues? :D

yaleman (changes requested):

webauthn-authenticator-rs/Cargo.toml
Firstyear: If a dep is in both build and dependencies, it'll merge the version to the parent in dependencies, so the rule of 0.10.41 from deps will apply here so I think leaving this as 0.10 is fine.

Firstyear (approved):

yaleman (approved):

Issue #252, opened 2023-01-07
Struggling to get Javascript example working


66OJ66:
Hi, I'm trying to create a basic Javascript example to work with the Axum server example.

I'm using these 2 files:

`index.html`
``` html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width, initial-scale=1">
    <title>WebAuthn-rs Tutorial</title>
    <script src="/assets/auth.js" async></script>
</head>
<body>
    <p>Welcome to the WebAuthn Server!</p>

    <input type="text" id="username" placeholder="Enter your username here">
    <button onclick="register()">Register</button>
</body>
</html>
```

`auth.js`
``` javascript
// Registers a user
function register () {
    let username = document.getElementById('username').value;
    if (username === "") {
        alert("Please enter a username");
        return;
    }

    fetch('http://localhost:8000/register_start/' + username, {
       method: 'POST'
    })
    .then(response => response.json() )
    .then(cco => {
        cco.publicKey.challenge = bufferDecode(cco.publicKey.challenge);
        cco.publicKey.user.id = bufferDecode(cco.publicKey.user.id);

        return navigator.credentials.create({
            publicKey: cco.publicKey
        });
    }).then((credential) => {
        console.log(credential);

        fetch('http://localhost:8000/register_finish', {
            method: 'POST',
            headers: {
                'Content-Type': 'application/json'
            },
            body: JSON.stringify({
                id: credential.id,
                rawId: bufferEncode(credential.rawId),
                type: credential.type,
                response: {
                    attestationObject: bufferEncode(credential.response.attestationObject),
                    clientDataJSON: bufferEncode(credential.response.clientDataJSON),
                },
            })
        })
    })
}

// Converts ArrayBuffer to URLBase64
function bufferEncode(value) {
    return btoa(String.fromCharCode.apply(null, new Uint8Array(value)));
}

// Converts Base64 to ArrayBuffer
function bufferDecode(value) {
    return Uint8Array.from(value);
}
```

I'm able to start registration with this successfully, but when I post to the finish_register endpoint I'm getting the `MismatchedChallenge` error (i.e. it's failing here: https://github.com/kanidm/webauthn-rs/blob/c76118185756f2d4aa8262b250601b1a01d12c52/webauthn-rs-core/src/core.rs#L379)

[...]

PR #258, opened 2023-01-11
Add a HTML/Javascript front-end example

66OJ66:
Hi, as requested on #252, this PR adds a basic HTML/Javascript front-end

Rather than just adding the `.html` and `.js` files to a new folder under _webauthn-rs/tutorial/frontend/javascript_, I've added 2 features (`wasm` and `javascript`) to the Axum example so it's easier to use either front-end. `wasm` is marked as a default feature. These features are mutually exclusive.

axum-extra is added as a dependency so the [SpaRouter](https://docs.rs/axum-extra/latest/axum_extra/routing/struct.SpaRouter.html) can be used. This also allowed me to simplify some of the code (e.g. remove `file.rs`).

The index_view() HTML has been moved into assets/wasm_index.html and I changed `/pkg/wasm_bg.wasm` to `/assets/wasm_bg.wasm` etc.
The rationale for this was I felt pkg doesn't really apply to HTML/Javascript, and I wanted to keep the directory names consistent. Feel free to revert this though.


Firstyear:
I'm happy with the rename of pkg to assets, but how is spaRouter actually accessing and serving the .wasm files? I can't see how it links to the built outputs. Am I missing something?

tutorial/server/axum/assets/auth.js
```diff
@@ -0,0 +1,93 @@
+function register () {
+    let username = document.getElementById('username').value;
+    if (username === "") {
+        alert("Please enter a username");
+        return;
+    }
+
+    fetch('http://localhost:8080/register_start/' + username, {
```
micolous, line 8: `username` [needs to be escaped](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent) here, and also below.

(it also looks like this is a bug with the Rust/WASM bindings, but that doesn't need to be addressed in this PR)

tutorial/server/axum/assets/js_index.html
```diff
@@ -0,0 +1,20 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <title>WebAuthn-rs Tutorial</title>
+	<script src="https://cdn.jsdelivr.net/npm/js-base64@3.7.4/base64.min.js"></script>
```
micolous, line 7: This file looks like it has a mix of tabs and spaces, so indentation is inconsistent

tutorial/server/axum/assets/auth.js
```diff
@@ -0,0 +1,93 @@
...
+                type: credential.type,
+                response: {
+                    attestationObject: Base64.fromUint8Array(new Uint8Array(credential.response.attestationObject), true),
+                    clientDataJSON: Base64.fromUint8Array(new Uint8Array(credential.response.clientDataJSON), true),
+                },
+            })
+        })
+        .then((response) => {
+            if (response.ok){
+                console.log("Registered!");
```
micolous, line 38: Can success / failure be shown on the page as well?

ie: `document.getElementById('status').innerText = "Registered!";`

tutorial/server/axum/src/main.rs
```diff
@@ -26,39 +20,15 @@ extern crate tracing;
 mod auth;
 mod startup;
 
+#[cfg(all(feature = "javascript", feature = "wasm"))]
```
micolous, line 23: What happens when neither feature is selected?

tutorial/server/axum/assets/js_index.html
micolous, line 7: As this pins a specific version of the package, can this use [subresource integrity](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) and `crossorigin="anonymous"`

tutorial/server/axum/README.md
```diff
@@ -0,0 +1,10 @@
+## Axum Server
+
+This demonstrates using Axum as the backend.
+
+By default, it serves the WASM front-end ([located here](https://github.com/kanidm/webauthn-rs/tree/master/tutorial/wasm "located here")). You will need to build this yourself first then copy the output into the assets directory.
+
+If you want to use the HTML/Javascript front-end instead, run the following command
+```rust
```
micolous, line 8: This is `sh`, not `rust`.

66OJ66:
@Firstyear So the wasm_index.html file contains this:
```html
<script type="module">
        import init, { run_app } from './assets/wasm.js';
        async function main() {
           await init('./assets/wasm_bg.wasm');
           run_app();
        }
        main()
</script>
```

With the server running, when you navigate to **localhost:8080/** it:
1. Imports functions from /assets/wasm.js
2. Runs the main() function which retrieves the .wasm file from /assets


So you see this
WASM Loading

then this
WASM Loaded

tutorial/server/axum/src/main.rs
66OJ66: In that case, nothing would be served when navigating to **localhost:8080/** - that's why I marked the `wasm` feature as default

tutorial/server/axum/README.md
66OJ66: Oops - corrected now

Firstyear:
> @Firstyear So the wasm_index.html file contains this:

I meant more that I can't see what serves the *files* in the code here. You've created an "/assets" folder but I can't see it. Additionally, you have tutorial/server/axum/assets/wasm_index.html but wasm_bg.wasm isn't in that folder either? How does the axum server get to and serve the wasm_bg.wasm file itself?

66OJ66:
> How does the axum server get to and serve the wasm_bg.wasm file itself?

Ah sorry - initially the WASM files had to be copied across manually into _tutorial/server/axum/assets_.

I've just added a new shell script to build and copy the WASM files across automatically (it's essentially the same as the existing `build.sh` script - I've just changed the input and output directories).

Firstyear (approved):

Firstyear:
@66OJ66 Thank you!

Issue #214, opened 2022-10-11
webauthn-authenticator-rs: build bindings from Transport/Token -> AuthenticatorBackend


micolous:
Context: https://github.com/kanidm/webauthn-rs/blob/master/designs/authenticator-library.md#implement-an-authenticatorbackend-for-transporttoken

This will:

* let the library work on platforms _without_ a platform-level WebAuthn API (eg: Linux)
* test the robustness of the CBOR implementation with a "complete" registration and authentication flow
* potentially allow us to replace Mozilla's `authenticator-rs` library (which only supports USB)

micolous:
> * Handle devices being plugged in / appearing after the prompt flow has started
>   * make `Transport.tokens()` async and return a `Stream` of `Token`?

I think this is should be the first thing to look at in terms of improving the user experience.  It would also unlock being able to handle authenticator selection.

Looking at device monitoring (hot-plug events) this on a transport level:

### BTLE

This should be possible with [`Adapter::events()`][0]. The library already uses something similar for `cable`; and it may need to be merged together.

[0]: https://docs.rs/btleplug/latest/btleplug/platform/struct.Adapter.html#method.events

### caBLE

The caBLE UI flow works differently, and already has to listen for BTLE advertisements to start the handshake process.

### NFC

The library already tracks reader states, and we can listen for events with `Context::get_status_change`.

### USB HID

It looks like `hidapi-rs` won't do this. You can only poll for devices with `HidApi::refresh_devices()` and `HidApi::device_list()`. Monitoring for device connection / disconnection also isn't supported by the underlying `hidapi` C library, so would require basically rewriting `hidapi-rs` (to eliminate the `hidapi` dependency) to make it work, but [there is a plan to do this upstream][1].

[1]: https://github.com/ruabmbua/hidapi-rs/projects/1

`rusb` looked promising, but it looks like it's using `libusb` underneath, so HID device access will be problematic.

Mozilla's `authenticator-rs` has its own HID handling code, which is Rust-native and supports monitoring for devices.  This tries to avoid taking dependencies on platform-specific crates, and instead provides its own bindings for everything.

I'm leaning towards forking Mozilla's `authenticator-rs` HID handling code; it seems to be the most complete here, but it will mean that we'd need to ship a pile of platform-specific code for it... which is something I had explicitly hoped to avoid.

micolous:
[...]

PR #272, opened 2023-02-05
BTLE authenticator transport

micolous:
This adds basic support for using Bluetooth Low Energy authenticators.

Because of API design and UI problems (see #214), the interface is *extremely* flaky and timing sensitive.  Fixing that will be a lot more work (refactoring more of the library to be `async`, and use `tokio` `Streams`); but should be easier to approach building around a transport which has the same model.

However, the important part of actually being able to use BTLE GATT as a transport layer, and being able to frame requests properly with CTAP's variable MTU works.

Tested with:

* Hideez Key 4
* Feitian MultiPass K16

Fixes #

### DIFF
--- a/webauthn-authenticator-rs/src/lib.rs
+++ b/webauthn-authenticator-rs/src/lib.rs
@@ -35,6 +35,7 @@
 //! the appropriate `--features` flag listed inline.
 //!
 //! * [CTAP 2.0, 2.1 and 2.1-PRE protocol implementation][crate::ctap2]
+//! * [Bluetooth][] (with `--features bluetooth`)
 //! * [caBLE][] (with `--features cable`)
 //! * [NFC][] via PC/SC API (with `--features nfc`)
 //! * [SoftPasskey][] (for testing)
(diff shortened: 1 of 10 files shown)


webauthn-authenticator-rs/src/lib.rs
```diff
@@ -35,15 +35,17 @@
...
 //!
 //! * [CTAP 2.0, 2.1 and 2.1-PRE protocol implementation][crate::ctap2]
+//! * [Bluetooth][] (with `--features bluetooth`)
 //! * [caBLE][] (with `--features cable`)
-//! * [Mozilla Authenticator][] (with `--features u2fhid`)
 //! * [NFC][] via PC/SC API (with `--features nfc`)
 //! * [SoftPasskey][] (for testing)
 //! * [SoftToken][] (for testing)
 //! * [USB HID][] (with `--features usb`)
+//! * [Mozilla Authenticator][] (with `--features u2fhid`)
```
Firstyear, line 44: Given how far the support in usb hid and other transports has come, we could consider removing this later.

micolous: That's something for later that I've previously noted in #214. While our own CTAP2 protocol support is _much_ better, there's usability/UI issues which block it being a true replacement.

PR #276, opened 2023-02-07
first pass at porting to boringssl

doramatadora:


Firstyear:
Hi there, we aren't interested to swap away from openssl to boringssl. We have requirements that make it necessary. boringssl could be an alternate feature version however, but I think it's something that should be raised as an issue an discussed first to work out what your requirements are. Thanks!

micolous:
I'll also mention the warning in [BoringSSL's readme][0]:

> BoringSSL is a fork of OpenSSL that is designed to meet Google's needs.
>
> Although BoringSSL is an open source project, it is not intended for general use, as OpenSSL is. **We don't recommend that third parties depend upon it.** (emphasis added) Doing so is likely to be frustrating because there are no guarantees of API or ABI stability.

While it looks like this uses Cloudflare's Rust bindings (that pin a specific version of BoringSSL), that's [currently][1] a [commit of BoringSSL from September 2021][2] (see also https://github.com/cloudflare/boring/issues/100).

There are *also* BoringSSL Rust bindings from BoringSSL itself ([bssl-sys][3]) which are a slightly newer target, but still subject to the same issue of BoringSSL having no stable API or proper "releases".

BoringSSL is an artefact of how Google internally manages third-party software which just happens to be exported publicly. It's not intended as "because it's from Google it must be better than OpenSSL" for non-Google projects.

[0]: https://boringssl.googlesource.com/boringssl/#boringssl
[1]: https://github.com/cloudflare/boring/tree/3059ba6e102599f8ae0a962223ca1e216bb61902/boring-sys/deps
[2]: https://github.com/google/boringssl/tree/f1c75347daa2ea81a941e953f2263e0a4d970c8d
[3]: https://boringssl.googlesource.com/boringssl/+/refs/heads/master/rust/

PR #284, opened 2023-02-21
Resolve potential client side UV bypass in passkeys

### LINKED ISSUE #281: Can't use Passkeys registered on devices with preferred authorization on devices without preferred authorization
Hi there, I'm running into an issue with using Passkeys between devices. Here's a quick way to reproduce it:

* Use the default `UserVerificationPolicy` of `Preferred`
* Register a Passkey on a device where preferred causes user verification to happen (such as an iPhone with Face ID)
* Attempt to use the Passkey on a different device where preferred doesn't trigger user verification (such as a Mac Studio)
* Get a `UserNotVerified` error

It seems like it's caused by this logic:
https://github.com/kanidm/webauthn-rs/blob/09ff14cec763863cbe2122a48fa72e36efc24e09/webauthn-rs-core/src/core.rs#L757-L764

As far as I can tell, this causes the passkey to get "upgraded" into a state where it can only be used with user verification. However, it seems to be difficult to work around this behavior.

My current solution is enabling the "danger-credential-internals" feature, checking if the passkey `Credential` has `user_verified` set to true, and if so, setting the user verification of the `RequestChallengeResponse` to `Required`. Does this make sense, is there anything else I should be doing instead?

Thanks!

Firstyear:
Fixes #281 In response to this issue we should take proactive steps to prevent client side UV bypassing in our apis. Since firefox is soon to gain CTAP2 support, being the last limiting factor for this, I believe it's time to enforce required UV in passkey flows. Since this won't be released for a bit until some other library changes occur, we can wait for ff to support this by default. 

### DIFF
--- a/webauthn-rs/src/lib.rs
+++ b/webauthn-rs/src/lib.rs
@@ -427,7 +427,7 @@ impl Webauthn {
         let credential_algorithms = self.algorithms.clone();
         let require_resident_key = false;
         let authenticator_attachment = None;
-        let policy = Some(UserVerificationPolicy::Preferred);
+        let policy = Some(UserVerificationPolicy::Required);
         let reject_passkeys = false;
 
         let extensions = Some(RequestRegistrationExtensions {
@@ -502,7 +502,7 @@ impl Webauthn {
     ) -> WebauthnResult<(RequestChallengeResponse, PasskeyAuthentication)> {
         let extensions = None;
         let creds = creds.iter().map(|sk| sk.cred.clone()).collect();
-        let policy = UserVerificationPolicy::Preferred;
+        let policy = UserVerificationPolicy::Required;
         let allow_backup_eligible_upgrade = true;
 
         self.core


yaleman (approved):

Firstyear:
Okay, I had a chat to @micolous and we decided that since this actually might break already registered credentials such as u2f, then we have to wait to see what apple does about their UV bypass issues in passkeys. If they fix it, then we can leave this as is. If they don't fix it, then we'll need to enforce this to work around the security hole.

micolous (changes requested):
Per discussion, lets wait for Apple on this one, as this will be a breaking change for some users, especially for U2F-only keys.

This will need documentation, ideally *before* this gets pushed out so that library users have a chance to plan for it.

This should have an escape hatch of some kind as well (feature flag?), particularly as even if Firefox stable gets FIDO2 support on non-Windows platforms, there's still the LTS branch which is carried by a bunch of Linux distributions.

Firstyear:
In a mail to the webauthn mailing list, the chrome team have now confirmed they have altered their uv preferred behaviour to match safaris allowing the "peripheral exchange bypass".

This means that for libraries outside of webauthn-rs, uv=preferred is once again bypassable, but for us it means there are potential inconsistencies in user experience and behaviour.

I think we need to make uv=required the default as a result of this since we now have two browsers that are susceptible to peripheral change affecting uv status.

Firstyear:
Mail description btw

> Starting with [Chrome 113](https://chromiumdash.appspot.com/schedule), Chrome’s macOS platform authenticator will no longer prompt for either factor if the Relying Party sets the [User Verification Requirement](https://www.w3.org/TR/webauthn-2/#enum-userVerificationRequirement) for the request to “preferred” or “discouraged” and Touch ID is unavailable. If the User Verification Requirement is set to “required”, Chrome continues to enforce user verification as before, and falls back to prompting for the device password if necessary. The [UV bit](https://www.w3.org/TR/webauthn-2/#flags) in the response will faithfully reflect whether user verification was performed or not.
> 
> This change minimizes user disruption flows in situations where Touch ID cannot be used for user verification, and makes Chrome’s implementation align more closely with Safari’s.

micolous (approved):
Approving because of Chrome's bypass behaviour.

Issue #288, opened 2023-03-12
Making requests from a Chrome extension


Alacho2:
**Is your feature request related to a problem? Please describe.**
I am trying to implement a backend server that uses this library to provide authentication from one of my projects.
The solution I have implemented works in the context of localhost, so I know I am on the right track. But I am now trying to port it to the frontend solution I am developing, an extension for Chrome. 

```
let rp_id = "jnpfkofnigkaocfcdcdppaokjkmhjcio";
let rp_origin = Url::parse("chrome-extension://jnpfkofnigkaocfcdcdppaokjkmhjcio")
      .expect("Invalid url there, buddy");
```

This seems to manage to prompt everything that I need; however, when performing start_registration and finish_registration, I receive the following error: 
`The clients relying party id hash does not match the hash of our relying party id.`

I assume that this is me holding the library wrong somehow but could you shed some light on how this should be implemented?

Looking at the [Chromium code](https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/webauthn/chrome_authenticator_request_delegate.cc;l=198-206;drc=de68be3f18ba99cc01d75903e167ca09bade253c) for this section tells me that this is the way to do it. 

Any input would be appreciated.

Firstyear:
If you're getting invalid rpid hash that means you're hitting this:

https://github.com/kanidm/webauthn-rs/blob/master/webauthn-rs-core/src/core.rs#L427

So you'll likely need to change the rp_id you are using in your builder, but I also note https://github.com/kanidm/webauthn-rs/blob/master/webauthn-rs/src/lib.rs#L202 which checks for effective origin.

So I think we need to see what you are doing with the webauthn builder here or your calls to rs-core to help diagnose, but most likely your rp-id isn't set correctly *or* we aren't able to handle that style of chrome url.

yaleman:
Dear github, I said refers, not fixes!

Alacho2:
Thanks for coming back to me.

I put up a [barebone example](https://github.com/Alacho2/barebone-webauthn) of how I use the library, based on the tutorial belonging to your repo. Provided that you change the id defined in main.rs after you've loaded the extension in chrome://extensions. Might be easier for you guys to debug this than it is for me.

Firstyear:
Thanks, I'm at a conference right now but I'll try to debug this during the week or next week :)

Firstyear:
```
[...]

PR #289, opened 2023-03-13
making a basic test for a webauthn object from a chrome extension

yaleman:
Refers to #288

### DIFF
--- a/webauthn-rs-core/src/core.rs
+++ b/webauthn-rs-core/src/core.rs
@@ -50,10 +50,6 @@ use base64urlsafedata::Base64UrlSafeData;
 /// as much as possible.
 #[derive(Debug)]
 pub struct WebauthnCore {
-    /// The RP name
-    #[cfg(debug_assertions)]
-    pub rp_name: String,
-    #[cfg(not(debug_assertions))]
     rp_name: String,
     rp_id: String,
     rp_id_hash: [u8; 32],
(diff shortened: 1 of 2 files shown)


webauthn-rs-core/src/core.rs
```diff
@@ -50,6 +50,10 @@ use base64urlsafedata::Base64UrlSafeData;
 /// as much as possible.
 #[derive(Debug)]
 pub struct WebauthnCore {
+    /// The RP name
+    #[cfg(debug_assertions)]
+    pub rp_name: String,
```
Firstyear, line 55: Could just be worth making a getter instead of the debug assertions.

webauthn-rs/src/lib.rs
```diff
@@ -1225,3 +1225,16 @@ impl Webauthn {
         self.core.authenticate_credential(reg, &state.ast)
     }
 }
+
+#[test]
+/// Test building a webauthn object from a chrome extension origin
+fn test_webauthnbuilder_chrome_url() {
```
Firstyear, line 1231: does the test pass or fail? 🤔

webauthn-rs-core/src/core.rs
yaleman: done!

webauthn-rs/src/lib.rs
yaleman: it passes! and now the comment mentions it.

Firstyear (approved):

PR #285, opened 2023-02-22
Add design philosophy doc

micolous:
This has come up a few times; #277 is just the most recent time :)

Fixes #

- [x] cargo test has been run and passes
- [x] documentation has been updated with relevant examples (if relevant)

PR #292, opened 2023-03-28
Add residentKey to AuthenticatorSelectionCriteria

conr2d:
`AuthenticatorSelectionCriteria` has the field of [residentKey](https://www.w3.org/TR/webauthn-2/#dom-authenticatorselectioncriteria-residentkey), but webauthn-rs doesn't support it.

`residentKey` can have either `discouraged`, `preferred` or `required`.

When its value is not present, `requireResidentKey` is used, but we don't have the way to handle the `preferred` case.

|residnetKey|requireResidentKey||
|-|-|-|
|-|false|"discouraged"|
|-|true|"required"|
|?|?|"preferred"|

This PR breaks the API, so please let me know which way is preferred between breaking the existing or adding a new one. (or your decision that `won't fix` this)


Firstyear:
I'm going to review and think about this more tomorrow, but the problem is preferred is identical to required in all cases in all browsers.

Firstyear:
So re-reading the spec it seems I missed in the update to l2 this section https://www.w3.org/TR/webauthn/#dictionary-authenticatorSelection 

We support the legacy "L1" version of this which is the bool, whereas the "L2" and above are using the enum.

I think that currently due to the attitude of the browsers, and the fact that required and preferred are identical as options, that the "bool" of require resident key is still correct here as the method for an RP to request if they want an RK or not, since that's exactly how a browser will treat it. 

Saying this, we *should* be supporting the values listed in L2. 

So I would like the pr change such that:

* The call to generate_challenge_register_options is still the same as the current api (with there being a bool of require / not require)
* If that is set to required (true) then we set both the bool of require_resident_key *and* we set the resident key value to "required".
* If that value is set to false, then we set require resident key to false, and resident key to "discouraged".

In the future if we wish to expose the actual details of what the rk setting can do we can do so then with the breaking api change. But right now, I think key residence is a really problematic part of the spec, and preferred and required are identical so exposing this as a false dichotomy doesn't sit right with me. @micolous what do you think?

micolous:
> @micolous what do you think?

I think you've covered the issue pretty well there.

On a CTAP level, there's only a boolean `rk` option in the request, but I guess the "preferred" behaviour is that you'd check the authenticator's `GetInfo` first and adjust things.

The other fun part is dealing with Windows (which blocks direct CTAP access to authenticators) - `WEBAUTHN_AUTHENTICATOR_MAKE_CREDENTIAL` has [BOOL bRequireResidentKey](https://github.com/microsoft/webauthn/blob/9fb38aa2775e286f504fc087efc62e5557bee881/webauthn.h#L540-L541), and only V4 of that struct adds then [BOOL bPreferResidentKey](https://github.com/microsoft/webauthn/blob/9fb38aa2775e286f504fc087efc62e5557bee881/webauthn.h#L579-L581).  V4 of the struct is only in WebAuthN API v3, which is only available on Windows 11. Windows 10 targets WebAuthn API v2 (and looks like MS aren't updating it!), which has `WEBAUTHN_AUTHENTICATOR_MAKE_CREDENTIAL_OPTIONS` V3... so even if "preferred" had any meaning, it would be impossible to implement in Windows 10.

Firstyear:
Well, the other side of this is also that there is no true signal if an rk was actually created or not outside of credProps. So generally even the bool is a "pls sir can I have some residency". 

And in the browser preferred/required case, it's still just a check over ctap1 vs all else. And even in these cases, credProps is the only way to check anyway. It's all a glorious hot mess. 

So I think having ambiguous "maybe rk, maybe not" isn't right. You either want it - or you don't. So having required (plus cred props) vs discouraged I think is the right distinction. We've tried to argue at the spec/browser level that rk preferred means "only create an rk if storage is unlimited" but that went no where, so we are stuck with this boolean.

micolous:
If browsers implemented `rk=preferred` the way the spec indicates[^1], an RP using it would need flows to handle non-resident keys, and could not rely on any properties of resident keys (such as usernameless authenticator selection)... so they could just as well use `rk=discouraged` and get exactly the same thing (with the bonus of [not consuming users' limited key storage][0]).

Going into the `webauthn-rs` side of things, it's a pretty opinionated API (see #285), and it has avoided implementing parts of the WebAuthn spec which are problematic, confusing, or otherwise provide no clear benefit. 😄 

[^1]: which they don't, and can't, because of [Windows 10][1]

[0]: https://fy.blackhats.net.au/blog/html/2023/02/02/how_hype_will_turn_your_security_key_into_junk.html
[1]: https://github.com/kanidm/webauthn-rs/pull/292#issuecomment-1487971040

Firstyear:
> If browsers implemented `rk=preferred` the way the spec indicates[1](#user-content-fn-1-bd81eb0dad623d2c0bd7775688d07a9d), an RP using it would need flows to handle non-resident keys, and could not rely on any properties of resident keys (such as usernameless authenticator selection)... so they could just as well use `rk=discouraged` and get exactly the same thing (with the bonus of [not consuming users' limited key storage](https://fy.blackhats.net.au/blog/html/2023/02/02/how_hype_will_turn_your_security_key_into_junk.html)).

Actually, even with rk=required they have to support alternate flows because rk=required isn't actually actually enforced anyway because reasons. rk=required is and credProps is only the browser saying "yep I really did ask for a resident key" but there is actually no true indication one was even created at all.

webauthn-rs-proto/src/options.rs
```diff
@@ -197,6 +197,20 @@ pub enum AuthenticatorAttachment {
...
+/// The Relying Party's requirements for client-side discoverable credentials.
+///
+/// <https://www.w3.org/TR/webauthn-2/#enumdef-residentkeyrequirement>
+#[derive(Clone, Copy, Debug, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord)]
+#[serde(rename_all = "lowercase")]
+pub enum ResidentKeyRequirement {
+    /// <https://www.w3.org/TR/webauthn-2/#dom-residentkeyrequirement-discouraged>
+    Discouraged,
+    /// <https://www.w3.org/TR/webauthn-2/#dom-residentkeyrequirement-preferred>
+    Preferred,
```
Firstyear, line 209: Please add a comments warning like
"⚠️ In all major browsers preferred is identical in behaviour to required. You should use required instead."

Firstyear (changes requested):
One minor change otherwise looks good thank you!

Firstyear (approved):

Firstyear:
@micolous Are you happy with this?

micolous (approved):

micolous:
> @micolous Are you happy with this?

Looks good, thanks!

Firstyear:
Ps: @conr2d feel free to make an extra PR to add yourself to contributors.md if you so wish :)

PR #293, opened 2023-03-29
clippy: fix multiple issues

micolous:
Fixes multiple clippy failures in CI:

* (error) `derivable_impls` for many `Default` implementations (added in new clippy)
* (error) `expect_used` regressions missed in #289
* (error) mismatched `image` library versions from `bardecoder`
* (warning) `wrong_self_convention` in `Eid::to_bytes()`
* (warning) `bool_assert_comparison` in `nfc/atr.rs` tests
* (warning) `box_default`  in multiple places

- [x] cargo test has been run and passes
- [x] documentation has been updated with relevant examples (if relevant)

PR #294, opened 2023-03-29
Fix CI failure in clippy check

conr2d:
CI check in #292 fails because of clippy errors that are fixed by this PR. 


conr2d:
These issues are already covered by #293.

PR #300, opened 2023-04-12
Fix broken deviceplug commit due to force-push of dev branch

micolous:
We were pinning a commit of `deviceplug` for a Linux-specific bug fix; but this was only on the upstream `dev` branch.

That commit got stomped due to a force-push of that branch: https://github.com/deviceplug/btleplug/pull/302#issuecomment-1503946059


AtunCL:
@micolous how can i contact you for a question about proxmark3 and rfid mifare plus 2k sl0/sl3

Firstyear (approved):

PR #302, opened 2023-04-14
Add remaining FIDO 2.1 GetInfoResponse fields

micolous:
This also adds a `Default` impl for `GetInfoResponse`, and makes the Display output a little less verbose when options are missing.

This PR doesn't conflict with others I have in flight, so ordering doesn't matter.


Firstyear (approved):
Looks great!

PR #303, opened 2023-04-16
20230416 passkey define

Firstyear:
This generally improves docs around passkeys and webauthn-rs generally, but also improves the docs around the uv preference options that we recently changed. 

### DIFF
--- a/webauthn-rs-proto/src/options.rs
+++ b/webauthn-rs-proto/src/options.rs
@@ -28,37 +28,42 @@ pub type CredentialID = Base64UrlSafeData;
 /// verification state of the credential in the persisted credential. These persisted
 /// credentials define which UserVerificationPolicy is issued during authentications.
 ///
-/// **IMPORTANT** - Due to limitations in the webauthn specification, CTAP devices, and browser
+/// **IMPORTANT** - Due to limitations of the webauthn specification, CTAP devices, and browser
 /// implementations the only secure choice is *required*.
 ///
-/// > ⚠️  **WARNING** - discouraged is marked with a warning, as in some cases, some authenticators
-/// > will FORCE verification during registration but NOT during authentication. This means
-/// > that is is NOT possible assert verification has been bypassed or not from the server
-/// > viewpoint, and to the user it may create confusion about when verification is or is
-/// > not required leading users to not trust UV is effective.
+/// > ⚠️  **WARNING** - discouraged is marked with a warning, as some authenticators
+/// > will FORCE verification during registration but NOT during authentication.
+/// > This makes it impossible for a relying party to *consistently* enforce user verification,
+/// > which can confuse users and lead them to distrust user verification is being enforced.
 ///
 /// > ⚠️  **WARNING** - preferred can lead to authentication errors in some cases due to browser
-/// > peripheral exchange attacks. Webauthn RS is not vulnerable to these attacks due to our
+/// > peripheral exchange allowing authentication verification bypass. Webauthn RS is not vulnerable
+/// > to these bypasses due to our
 /// > tracking of UV during registration through authentication, however preferred can cause
 /// > legitimate credentials to not prompt for UV correctly due to browser perhipheral exchange
-/// > attacks leading Webauthn RS to deny them in what should otherwise be legitimate operations.
+/// > leading Webauthn RS to deny them in what should otherwise be legitimate operations.
 #[derive(Clone, Copy, Debug, Default, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord)]
 #[allow(non_camel_case_types)]
 #[serde(rename_all = "lowercase")]
 pub enum UserVerificationPolicy {
-    /// Require User Verification bit to be set, and fail the registration or authentication
-    /// if false. If the authenticator is not able to perform verification, it may not be
+    /// Require user verification bit to be set, and fail the registration or authentication
+    /// if false. If the authenticator is not able to perform verification, it will not be
     /// usable with this policy.
+    ///
+    /// This policy is the default as it is the only secure and consistent user verification option.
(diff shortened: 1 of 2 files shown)


webauthn-rs-proto/src/options.rs
```diff
@@ -28,24 +28,37 @@ pub type CredentialID = Base64UrlSafeData;
...
     Required,
-    /// TO FILL IN
+    /// Prefer uv if possible, but ignore if not present. In other webauthn deployments this is bypassable
+    /// as it implies the library will not check uv is set correctly for this credential. Webauthn-RS
+    /// is *not* vulnerable to this as we check the uv state always based on it's presence at registration.
     #[serde(rename = "preferred")]
-    #[default]
     Preferred,
-    /// TO FILL IN
+    /// Discourage - but do not prevent - userverification from being supplied. Many CTAP devices
```
micolous, line 60: `Many CTAP devices` => `CTAP 2 devices with a PIN set`

It's hard to describe this as "default" or not, as CTAP 2.0 and 2.1 do different things, the "default" state of `alwaysUv` and `makeCredUvNotRqd` are not defined, and it depends which version of CTAP both the authenticator *and* initiator are using.

For a device with PIN/UV authentication configured, using _non-discoverable credentials_, where both authenticator and initiator are using the same CTAP version:

CTAP version | MakeCredential | GetAssertion
---- | --- | ----
CTAP 2.0 | auth required | auth optional
CTAP 2.1 with `alwaysUv = true` | auth required | auth required
CTAP 2.1 with `makeCredUvNotRqd = true` | auth optional | auth optional
CTAP 2.1 otherwise | auth required | auth optional

webauthn-rs-proto/src/options.rs
```diff
@@ -28,24 +28,37 @@ pub type CredentialID = Base64UrlSafeData;
...
+/// implementations the only secure choice is *required*.
+///
+/// > ⚠️  **WARNING** - discouraged is marked with a warning, as in some cases, some authenticators
+/// > will FORCE verification during registration but NOT during authentication. This means
+/// > that is is NOT possible assert verification has been bypassed or not from the server
+/// > viewpoint, and to the user it may create confusion about when verification is or is
+/// > not required leading users to not trust UV is effective.
+///
+/// > ⚠️  **WARNING** - preferred can lead to authentication errors in some cases due to browser
+/// > peripheral exchange attacks. Webauthn RS is not vulnerable to these attacks due to our
```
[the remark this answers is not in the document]
Firstyear: The announcement was just via a mailing list so I'mu not sure we can reference it. But otherwise I've reworded this.

webauthn-rs-proto/src/options.rs
Firstyear: The problem is the 2.1 defaults. makeCredUvNotRqd is *not* set by any browser, it's only a ctap level thing for other applications. And alwaysUv is a perdevice setting and again, can't be controlled by an RP. So IMO we have to assume the default "out of box" of a key for 90% of deployments which is the default.

webauthn-rs-proto/src/options.rs
```diff
@@ -28,24 +28,37 @@ pub type CredentialID = Base64UrlSafeData;
 /// verification state of the credential in the persisted credential. These persisted
 /// credentials define which UserVerificationPolicy is issued during authentications.
 ///
-/// ⚠️  WARNING - discouraged is marked with a warning, as in some cases, some authenticators
-/// will FORCE verification during registration but NOT during authentication. This means
-/// that is is NOT possible assert verification has been bypassed or not from the server
-/// viewpoint, and to the user it may create confusion about when verification is or is
-/// not required.
+/// **IMPORTANT** - Due to limitations in the webauthn specification, CTAP devices, and browser
+/// implementations the only secure choice is *required*.
```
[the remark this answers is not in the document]
Firstyear: Yes, but this is where the UV on reg but not auth becomes a problem. I've done user surveys in the past where people who do a reg where UV is required will believe it's then required always going forward, and are surprised when it's not which leads them to think "oh UV is clearly just a weird flakey sometimes thing".

webauthn-rs-proto/src/options.rs
Firstyear: Worth saying with credProtect we can force uv always at least, but that doesn't fix the problem very nicely.

micolous: `alwaysUv` and `makeCredUvNotRqd` are authenticator configuration attributes in this context, but otherwise I think we agree about the issue.

I think that saying "Many CTAP devices" makes it sound like that it's at manufacturer's discretion, whereas ascribing it to the presence of a PIN is more accurate and more likely to have happened (eg: [initiators could try to configure on CTAP 2.x authenticators when `uv = required`][0]).

From memory, only the Yubikey Bio (which supports CTAP 2.1) set `alwaysUv = true` out of the box _and_ demanded you set a PIN (and had an annoying flashing LED when it was unconfigured), otherwise authenticators require explicit configuration _and_ CTAP 2.1 to force UV on GetAssertion.

There is the issue of resident keys (which work differently), but I think we can hand-wave that part for now.

[0]: https://fidoalliance.org/specs/fido-v2.1-ps-20210615/fido-client-to-authenticator-protocol-v2.1-ps-errata-20220621.html#sctn-makeCred-platf-actions

Firstyear: Most authenticators will prompt to setup a pin on uv=preferred too, so in the majority of cases a pin will exist on the device meaning this situation occurs.

Firstyear: And the bio is a special case since it's presence check is also a verification so I can see why they set alwaysUv on those.

micolous (approved):

PR #305, opened 2023-04-17
add CLI tool to derive caBLE tunnel server domain

micolous:


yaleman (approved):

yaleman (approved):

webauthn-authenticator-rs/Cargo.toml
```diff
@@ -91,3 +91,7 @@ image = ">= 0.23.14, < 0.24"
 
 [build-dependencies]
 openssl.workspace = true
+
```
Firstyear, line 94: I think we need to talk in future about how to move these out of being examples, since there are multiple tools here that actually are good cli tools, and not examples.

PR #310, opened 2023-04-19
Implement the basis of a device catalog

### LINKED ISSUE #238: Fido MDS tool query
We need a way to be able to query the FIDO MDS data, but more specifically, users who use the tool need a way to query it as well.

How should we do this? What should the queries look like? 

What kind of requirements would users realistically have? 

Finally, when we have the query complete, what format should we output the data as? Something we can import directly into Webauthn-RS Attestation CA Lists?

Currently some ideas in my head for use cases are:

* A user wants to allow device AAGUID X and Y and export these to an attestation CA List
* A user wants to know what devices have been certified with "FIPS-CMVP-2"> X.
* A user may want to know what devices support a specific extension such as hmac-secret.
* A user wants to know what devices support a specific userverification method - *or do not* support a verification method (exclude bio metrics for example)

My thinking is that users will often query to find devices, shortlist them, then can just feed the aaguid list to extract the set of attestation's CA's in use (and the aaguids they govern). 

Thoughts?

### LINKED ISSUE #307: Attestation in webauthn-rs
Allow attestation for passwordless credentials in webauthn-rs with a well defined mds/ca format.

Firstyear:
Fixes #307 Fixes #238 Fixes - This expands query support in the mds, adds our own device catalog, and allows converting out device catalog entries to attestation CA lists.

Still outstanding:

- [ ] improve query options for fido
- [x] Allow fido mds queries to convert to attestation ca lists


Firstyear:
I think my goal now is to just improve the query and output of fido mds first, get it merged, then come back to the device catalog.

Firstyear:
I would love some review on this before I continue, I think the next big thing will more query options generally. But getting out an attestation CA list is now possible.

Firstyear:
@micolous I think this is good to review. The minimal amounts of the fido query code is ready, and the device catalog is currently an internal build detail only.

PR #315, opened 2023-04-27
Migrate to clap 4.2, run binaries/examples with --help in CI, and clean up examples

### LINKED ISSUE #313: Re-enable clippy in CI for stable once clap issues addressed.
https://github.com/kanidm/webauthn-rs/pull/312 temporarily disabled Clippy in CI for 1.69 due to spuriously-failing clap correctness checks: https://github.com/rust-lang/rust-clippy/issues/10421

We'll need to revert that at some point once the upstream issue is fixed in a stable release.

There's an existing request to cherry-pick the fix in https://github.com/rust-lang/rust-clippy/issues/10685

micolous:
There was an issue with clippy throwing correctness errors on Rust 1.69 for clap 3.2.x's macros, so I had previously disabled it.  Migrating to clap 4.x fixes this, so we can re-enable clippy in CI.

clap 4.x has a few API changes, the main one for us was that it now the original struct field name as a field ID, rather than the "normalised" form (changing `_` to `-`), so `conflicts_with_all` and `ArgGroup`s need updating.  It also looks like some parts of clap only check argument definitions at runtime, and errors don't show up with `cargo test`; so we need to run those binaries manually with `--help` to discover errors.

While here, I've done some house-keeping on many `webauthn-authenticator-rs` examples:

* rename `authenticate/main.rs` to `authenticate.rs`
* rename `cable_domain/main.rs` to `cable_domain.rs`
* `cable_tunnel` example now uses `required-features` in `Cargo.toml`; so `cable_tunnel/main.rs` is gone, and effectively renaming `cable_tunnel/core.rs` to `cable_tunnel.rs`. This also fixes a broken link in the `cable-tunnel-server` README.
* delete obsolete `nfc_token_info` example, as `fido-key-manager` can do the same thing
* rename `softtoken/main.rs` to `softtoken.rs`

Fixes #313


fido-key-manager/src/main.rs
```diff
@@ -69,8 +69,8 @@ pub struct RenameFingerprintOpt {
 
 #[derive(Debug, Args)]
 pub struct RemoveFingerprintOpt {
-    /// The template ID
-    #[clap(min_values(1), required(true))]
+    /// The template ID(s) to remove
```
Firstyear, line 72: Does this actualy assert that 1 or more ids are provided.

Firstyear (approved):

fido-key-manager/src/main.rs
micolous: Yes:

```
> .\target\debug\fido-key-manager.exe remove-fingerprint
error: the following required arguments were not provided:
  <ID>...

Usage: fido-key-manager.exe remove-fingerprint <ID>...

For more information, try '--help'.
```

PR #324, opened 2023-06-06
Event-based Transport/Token API / USB HID rewrite

micolous:
This rewrites most of the Transport/Token API to be event based (#214), and in the process, replaces `hidapi` with a homebrew implementation. The branch is called `mozilla-hid` because my original plan was to base it on `authenticator-rs`, but in the end I've rewritten most of that.

I'm only targetting Linux, macOS and Windows for now.

Main changes:

* New USB HID API bindings (`fido-hid-rs`):
  * on Linux, this uses udev and the kernel hidraw interface
  * on macOS, this uses IOKit
  * on Windows, this uses the UWP HID API (available on Windows 10 and later); `hidapi` and `authenticator-rs` used the older Windows 2000 APIs which are not async friendly

* Add event-based API for device enumeration (`Transport::watch_tokens()`), moved Bluetooth and NFC transports to it.

  This is now the only way to enumerate tokens for Bluetooth.

* Started adding Bluetooth support to `fido-key-manager`. This is only really reliable on macOS at the moment.

* NFC: work around a bug where macOS' login manager tries to access the PIV applet on tokens *that we have an exclusive connection to*, which causes some tokens to reset (if the command arrives after `SELECT FIDO` but before other commands).

* Implement selection for CTAP 2.0/2.1-PRE authenticators

TODO list:

* [x] finish migrating fido-key-manager to the new API
* [x] finish migrating examples to the new API
* [x] consistently use tokio's threading/workers for USB backends rather than `std::thread`
* [x] figure out if the platform-specific USB stuff needs to be its own crate to handle optional platform-specific dependencies
* [x] clean up commented out code
* [x] check Bluetooth device filtering works correctly on all platforms
* [x] merge #328
* [x] merge #329 

----


Firstyear:
Looking really good <3

Firstyear:
This is getting pretty large, is there anything blocking it from merging? since we aren't planning to release just yet, did you want to rebase/commit and continue? What else do you think needs to be done?

micolous:
I think this is pretty close to done, but there are still some broken things that I'm working through on the checklist. I'll probably split off some of the smaller things which aren't _directly_ related to the rewrite into a separate PR; I just needed somewhere to stash notes for now.

Firstyear (approved):

PR #329, opened 2023-07-04
migrate from winres to embed-resource and fix cross-compiling for Windows

micolous:
Split from #324 

This should also fix cross-compiling with mingw toolchain.


fido-key-manager/Cargo.toml
```diff
@@ -28,5 +28,5 @@ tracing.workspace = true
 tracing-subscriber.workspace = true
 webauthn-rs-core.workspace = true
 
-[target.'cfg(windows)'.build-dependencies]
-winres = "0.1"
+[build-dependencies]
```
yaleman, line 31: Since this is windows-specific, shouldn't it still be limited thus?

yaleman (approved):

fido-key-manager/Cargo.toml
Firstyear: cargo.toml doesn'tu have a concept of OS though for deps.

yaleman: oh that's right, cross-compile. sorted! 😄

yaleman (approved):

PR #336, opened 2023-07-25
SessionLayer::with_secure(false)

paeifbnaeufbpae:
Enables the tutorial to be runnable locally by default. Without this, `finish_register` returns `Err(WebauthnError::CorruptSession)` when running locally.

### DIFF
--- a/tutorial/server/axum/src/main.rs
+++ b/tutorial/server/axum/src/main.rs
@@ -43,7 +43,7 @@ async fn main() {
     let session_layer = SessionLayer::new(store, &secret)
         .with_cookie_name("webauthnrs")
         .with_same_site_policy(SameSite::Lax)
-        .with_secure(true);
+        .with_secure(false); // TODO: change this to true when running on an HTTPS/production server instead of locally
 
     // build our application with a route
     let app = Router::new()


yaleman:
This should realistically always be run on a HTTPS endpoint, and a secure default is in line with our design standards.

Firstyear:
I think the reason I had this as false was for http testing between a laptop and a nearby ipadh for passkey dev, but now I'm okay to change it since we have other options for testing.

Firstyear:
@paeifbnaeufbpae Can you elaborate on the reason or error you hit with this in your testing?

paeifbnaeufbpae:
On my setup (macOS 13.4.1, Safari 16.5.2), the session KV store isn't persisted between the different stages without this change. It's empty at the start of `fn start_register`, making the retrieval of key `"reg_state"` fail.

Firstyear:
http://localhost should be considered a secure origin. So there is something else going on. Does it work on chrome / firefox? I wonder if safari is breaking local storage or cookies here somehow.

yaleman:
@paeifbnaeufbpae is this still an issue?

muxcmux:
Works on Chrome, but Safari does not store the cookie and doesn't send it back with the `register_finish` request unless `SessionLayer::with_secure` is set to false. I'm on macos 13.5.1 and Safari 16.6 (18615.3.12.11.2)

Firstyear:
My guess is that safari does not consider localhost a secure origin, where chrome and ff does.

I don't see an issue adding this then.

Firstyear (approved):

PR #338, opened 2023-08-10
Add Clone derivation to Webauthn struct

davidcaseria:
Fixes #


Firstyear (approved):
I am not opposed to this, but is there also a reason to avoid an ARC? Since this is a long-lived structure, an ARC should be cheap to establish at the start of execution.

Firstyear:
Funny, because I actually just found a need for this to be clone (being a member of cow-cell in concread).

PR #342, opened 2023-08-19
migrating to serde_cbor_2

### LINKED ISSUE #341: Migrate to serde_cbor_2

Per the discussion and work in #186, migrate to `serde_cbor_2` which seems to have sprouted from that.

Update the dependencies and release a patched version 😄

yaleman:
Fixes #341 


Cargo.toml
```diff
@@ -56,21 +54,35 @@ hex = "0.4.3"
...
 http-body-util = "=0.1.0-rc.2"
-hyper = { version = "=1.0.0-rc.3", default-features = false, features = ["http1"] }
+hyper = { version = "=1.0.0-rc.3", default-features = false, features = [
+    "http1",
+] }
 nom = "7.1"
 openssl = "^0.10.41"
 rand = "0.8"
 serde = { version = "^1.0.141", features = ["derive"] }
-serde_cbor = { version = "0.12.0-dev", package = "serde_cbor_2" }
```
Firstyear, line 64: I'm confused why we need to do this? We already use serde_cbor_2 with this line, is there a reason to actually change it over?

yaleman: Because cargo audit flips, and... it's one of those least-surprise things IMO

Firstyear: Sounds like a bug in cargo-audit :( I'll report it, but I'll accept this as well then.

Firstyear (approved):

PR #345, opened 2023-08-27
Update base64urlsafe version flag

Firstyear:
Update version of base64urlsafe to allow vendoring from git in Kanidm, update workspace resolver.


Firstyear:
Actually this breaks things.

PR #348, opened 2023-08-31
20230206 238 fido mds and quirks

### LINKED ISSUE #238: Fido MDS tool query

### LINKED ISSUE #307: Attestation in webauthn-rs

Firstyear:
Fixes #307 Fixes #238 Fixes - This expands query support in the mds, adds our own device catalog, and allows converting out device catalog entries to attestation CA lists.

Supersedes #310


Firstyear:
Example usage:

```
# fido-mds-tool fetch
2023-08-31T01:05:27.620536Z  INFO fido_mds_tool: Fetching from https://mds.fidoalliance.org/ to "/tmp/mds.blob.jwt"
2023-08-31T01:05:30.753665Z  INFO fido_mds_tool: Ok!

## Show all devices that are greater than or equal to l1 certification, and can perform fingerprint uv
# fido-mds-tool query "status gte l1 and uvm cnt fingerprint_internal"

## Using the same query, output the certificate roots and matching authenticator aaguids that conform to this query.
# fido-mds-tool query "status gte l1 and uvm cnt fingerprint_internal" -o
 
```

attestation-ca/src/lib.rs
```diff
@@ -0,0 +1,194 @@
...
+    fn try_from(iter: &[(&[u8], Uuid)]) -> Result<Self, Self::Error> {
+        let mut cas = BTreeMap::default();
+
+        for (der, aaguid) in iter {
+            let ca = x509::X509::from_der(der)?;
+
+            let kid = ca
+                .digest(hash::MessageDigest::sha256())
+                // Is there a better way to do this with fromIterator?
+                .unwrap();
```
micolous, line 135: Propagate error rather than unwrap, this looks like it returns `OpenSSLErrorStack` anyway

device-catalog/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+description = "Webauthn RS Device Catalog"
+repository = "https://github.com/kanidm/webauthn-rs"
+readme = "README.md"
+keywords = ["webauthn", "authentication"]
+categories = ["authentication", "web-programming"]
+license = "MPL-2.0"
+
+[dependencies]
+webauthn-attestation-ca.workspace = true
+base64urlsafedata = { version = "0.1.2", path = "../base64urlsafedata" }
```
micolous, line 15: Use workspace versions of these too

device-catalog/src/certificate_authority.rs
```diff
@@ -0,0 +1,28 @@
...
+
+use std::fmt;
+
+pub struct Authority {
+    pub ca: x509::X509,
+}
+
+impl fmt::Debug for Authority {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        /*
```
micolous, line 13: Delete commented code?

device-catalog/src/data/mod.rs
```diff
@@ -0,0 +1,6 @@
+pub mod android;
```
micolous, line 1: Consider using `include_bytes!` for certificates, and documenting their sources here (as `rustdoc`).

That way all certificates are normal files, rather than embedded in code, and it's very clear it comes from a third-party source.

micolous (approved):
Looks good to me; you'll want to rebase to pick up #349.

Thanks :)

PR #353, opened 2023-09-28
352 serde bincode

### LINKED ISSUE #352: Bincode serialization
### I did this
I tried to serialize and then deserialize a passkey using bincode

### I expected the following
Getting the original passkey back

### What actually happened
A runtime error from bincode, `DeserializeAnyNotSupported`

### Version (and git commit)
`0.5.0-dev`, `840a6f5cc7f65bd3c585ab00293addf76f70d973`

Linux

### Any other comments
This happens because `Base64UrlSafeData` uses `Deserializer::deserialize_any` in its deserializer but that isn't supported by bincode. I see that the serializer uses `serialize_str`. Is it possible to also just use `deserialize_str` instead of `deserialize_any`? Or, maybe even better, use `{serialize,deserialize}_bytes`?

Bincode is an important serialization format because it has the lowest memory footprint of all formats I know. If you want to store a lot of passkeys for a large user base, the memory footprint matters.

This issue applies not just to bincode, but to all serialization formats that don't serialize type information, e.g. also postcard and others. `deserialize_any` only works for verbose serialization formats that store the type. See [documentation](https://docs.rs/serde/1.0.188/serde/de/trait.Deserializer.html#tymethod.deserialize_any) Doesn't seem like that would be necessary for something like `Base64UrlSafeData`?

Firstyear:
Fixes #352 - This adds a hidden feature that allows the storage of vec<u8> to skip base64 for some users. 

We can't apply this into base64urlsafedata because features taint and having a flag to skip base64 would break wasm and json front ends. So as a result, the only way to proceed is to swap the type in the struct based on the cfg flag. It is a teeny bit messy because it introduces some useless into()/from() where without the feature we are doing base64 into base64, but it's needed when the feature is enabled. 

I can't think of a better way to actually achieve this though without breaking existing users horrendously. @micolous do you have any better ideas perhaps? 

### DIFF
--- a/base64urlsafedata/src/lib.rs
+++ b/base64urlsafedata/src/lib.rs
@@ -35,6 +35,12 @@ static ALLOWED_DECODING_FORMATS: &[GeneralPurpose] =
 /// when deserializing, will decode from many different types of base64 possible.
 pub struct Base64UrlSafeData(pub Vec<u8>);
 
+impl Base64UrlSafeData {
+    pub fn as_slice(&self) -> &[u8] {
+        self.0.as_slice()
+    }
+}
+
 impl fmt::Display for Base64UrlSafeData {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(f, "{}", URL_SAFE_NO_PAD.encode(self))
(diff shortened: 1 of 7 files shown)


yaleman:
looks like some of the tests are still salty about the redundant stuff..

Firstyear:
Yes, I wanted some input from @micolous first before I finish tidying that up

smessmer:
I believe an approach not relying on features is better. See the discussion [here](https://github.com/kanidm/webauthn-rs/issues/352#issuecomment-1740115516) in and some example code [here](https://gist.github.com/smessmer/5a022c8d09c764cf60e5fab3ad470141)

Firstyear:
Yeah. We think there are better ways.

PR #354, opened 2023-09-29
Add "HumanBinaryData" as an alternative to "Base64UrlSafeData"

micolous:
Related to issue #352, it **doesn't** fix things, but it's a start.

`HumanBinaryData` uses a more efficient "bytes" format when using non-human-readable serialisation formats (like CBOR).  The deserialiser still uses `deserialize_any` – I haven't got a great way around this yet.

I've done a few things differently to `Base64UrlSafeData`; for example, this doesn't implement `Display` or `FromStr` – they're different to what `Vec<u8>` does, and we should move away from this (this will be the focus of #356).

It also implements `Deref` and `DerefMut`, which should make the type more transparent – and all of the common impls are in macros.

I've added `bytes` support to `Base64UrlSafeData`, so it can read what `HumanBinaryData` produces. Unfortunately this makes migration a two-step process.

While we're here, this makes `serde_json` a dev-dependency of `base64urlsafedata`, because it's only used in tests; and this also adds a bunch more tests for expected input formats.


micolous:
Clippy appears to fail due to stricter checks on 1.70.0, possibly triggered by https://github.com/kanidm/webauthn-rs/pull/348

Firstyear (approved):
I think this is a great idea and a great way to approach this. Great work @micolous <3

PR #356, opened 2023-09-29
WIP: base64urlsafedata internals cleanup

micolous:
Goal is to remove all functionality from `Base64UrlSafeData` which is not about serialisation, and stop dependants from using internal-only interfaces. `HumanBinaryData` won't provide these interfaces, so it needs to make do.

* Make `Base64UrlSafeData.0` (the inner `Vec<u8>`) private, like `HumanBinaryData`.

  This switches over everything which used the inner field to using `From`, `Deref` or `::new()`. These interfaces are also available on `HumanBinaryData`, and makes the types pretty much interchangable.

* Remove `impl Display for Base64UrlSafeData` (which converted the inner value into a Base64 string).

  The alternative is to use `base64::Engine::encode`.

* Remove `impl TryFrom<&str> for Base64UrlSafeData` (which attempted to parse strings as Base64).

  The alternative is to use `base64::Engine::decode`.

* Migrate `WinWrapper::new` to use owned types rather than pass-by-reference.

  Most of the existing implementations made copies of that data, and callers don't need to clone anyway.

TODO:

* #363 


Firstyear (commented):
Should CredentialID in interface.rs in core become it's own type rather than relying on the CredentiaID from proto which requires b64? That way we can make this the humanReadableBinary type for serialisation.

micolous:
The difficulty is going to be migration. The serialisable `Credential` struct includes a `CredentialID` (which is currently `Base64UrlSafeData`); so if you serialise to/from a binary format, you'll need to get everything updated to something that can deserialise both (which #354 added).

We could remove it from parts of our API which do not get serialised – that could happen after this.

The goal is to get this into a state that we could swap it out with a `Vec<u8>` or a `HumanBinaryData` type, and making sure that everything which _actually needs_ Base64-related functionality (depending on it via `FromStr` or `Display`) instead explicitly encodes it as Base64.  This is one of the sources of confusion about the `Base64UrlSafeData` type.

Firstyear (commented):
What would be the conditions where we decide to commit this and proceed? currently my reading of this is that there are no incompatibilities with this change.

micolous:
Main thing left to check is the WASM/browser stuff

Firstyear:
We should wrap this up probably :) Do you want me to help?

micolous:
Closing in favour of #433

PR #367, opened 2023-10-11
Accept unknown transports in deserialisation

### LINKED ISSUE #364: Unknown transports
We have run into clients that send "cable" as a transport mechanism. This is not part of the webauthn spec but seems to happen in practice. Looking at the webauth spec, it says [relying parties must accept and ignore invalid transport values](https://www.w3.org/TR/webauthn-2/#dom-authenticatorattestationresponse-transports-slot). The current behavior of webauthn-rs is to crash instead because the credential can't be deserialized.

Deserialization of a `PublicKeyCredential` should not crash when unknown transports are present.

Firstyear:
Fixes #364


micolous:
Can this have a test for some unknown values?

Also this broke with `win10`: https://github.com/kanidm/webauthn-rs/actions/runs/6477577093/job/17587986234?pr=367#step:10:129

micolous (approved):

PR #371, opened 2023-10-18
Add feature for android specific work arounds

### LINKED ISSUE #365: Android Passkeys need changes to start_passkey_registration challenge
When running the challenge JSON returned by [start_passkey_registration](https://docs.rs/webauthn-rs/latest/webauthn_rs/struct.Webauthn.html#method.start_passkey_registration) against the Credential Manager API on Android, the device asks to set up a security key instead of a passkey. We are using webauthn-rs to create passkeys and are currently manually changing the challenge JSON to conform to what Android needs to set up a passkey:
```
authenticatorAttachment: "platform",
requireResidentKey: true,
userVerification: "required",
residentKey: "required",
```

Instead of us manually changing the output json from `start_passkey_registration`, it would be better if it either already had the correct values to create an Android passkey by default, or would allow us to modify these values via an `Options` input argument to `start_passkey_registration`.

Firstyear:
Fixes #365 


yaleman (approved):

webauthn-rs/src/lib.rs
```diff
@@ -493,6 +511,61 @@ impl Webauthn {
...
+    /// to pre-select this and send a correct set of options for either security key
+    /// *or* a synced google cloud key.
+    ///
+    /// If you choose to use this function you *MUST* ensure that the device you are
+    /// contacting is an Android device, and you *MUST* provide the user the choice
+    /// on your site ahead of time to choose between a security key / screen unlock
+    /// (triggered by [`start_passkey_registration`](Webauthn::start_passkey_registration))
+    /// or a synced google cloud key (triggered by this function).
+    #[cfg(feature = "workaround-android-specific-issues")]
+    pub fn start_android_only_passkey_registration(
```
micolous, line 529: This should be named something like "Google Passkey" rather than just "Android", because it is not "make passkey work on Android" but "allow Google Passkey stored in Google Password Manager".

This is also specific to Android devices with GMS (which covers the majority of devices, generally excluding those sold in PRC), just to make things more fun.

webauthn-rs/src/lib.rs
```diff
@@ -126,6 +126,24 @@
 //! the UV *may* occur on registration and then will not occur again, and that is *by design*.
 //!
 //! If in doubt, do not enable this feature.
+//!
+//! ## Android Specific Workarounds
+//!
+//! Android has a number of issues in the dialogs they present to users for authenticator
+//! selection. Instead of allowing the user to choose what kind of passkey they want to
+//! use and create (security key, device screen unlock, sync google cloud key), Android
```
micolous, line 134: `s/google cloud key/Google Password Manager-synchronised Passkey/`

"Google Cloud Key" may be confused with [GCP Key Management Service][0] or [Titan Security Key][1], when this is actually about [Google Password Manager][2].

[0]: https://cloud.google.com/security-key-management
[1]: https://cloud.google.com/titan-security-key
[2]: https://developers.google.com/identity/passkeys/supported-environments

micolous (changes requested):

micolous (approved):

PR #380, opened 2023-10-31
Minor changes for kani release

Firstyear:


yaleman (approved):

Firstyear:
@micolous We're gonna need to do something to get away from the git rev versions here so we can do releases in future I think.

attestation-ca/Cargo.toml
```diff
@@ -1,7 +1,13 @@
 [package]
 name = "webauthn-attestation-ca"
 version = "0.1.0"
+authors = ["William Brown <william@blackhats.net.au>"]
 edition = "2021"
+rust-version = "1.70.0"
+license = "MPL-2.0"
+description = "Webauthn Attestation Library"
+# documentation = "https://docs.rs/kanidm/latest/kanidm/"
+repository = "https://github.com/kanidm/webauthn-rs"
```
yaleman, line 10: These could inherit from the workspace

yaleman (approved):

Firstyear:
I'd like @micolous input before I proceed on this :)

micolous (approved):

micolous (changes requested):
Actually, this breaks on CI

fido-key-manager/src/main.rs
```diff
@@ -13,7 +13,7 @@ use std::time::Duration;
 use tokio_stream::StreamExt;
 #[cfg(feature = "solokey")]
 use webauthn_authenticator_rs::ctap2::SoloKeyAuthenticator;
-use webauthn_authenticator_rs::prelude::WebauthnCError;
+// use webauthn_authenticator_rs::prelude::WebauthnCError;
```
micolous, line 16: This should be gated on `#[cfg(feature = "solokey"]` to fix the warning.

Firstyear:
@micolous We had a bunch of issues generally, because we can't use git versions of things on crates.io, which meant btleplug was problematic. So we probably generally have a bunch of cleanup to get to a releasable state, which I want to do "sooner than later"

fido-key-manager/src/main.rs
Firstyear: K

Firstyear:
I dunno what I just did, but I think I cooked this PR 🤣

I think I close it in favour of another attempt

PR #400, opened 2023-12-15
20231215 264 cleanup

### LINKED ISSUE #264: Cleanup of core library
The webauth-rs-core library has "evolved" and changed shape a few times. This has led to the interfaces and code being "messy" (but not incorrect). In order to improve the user experience of the lower levels, and to make maintenance easier, we should clean this up.

Firstyear:
Fixes #264 - this cleans up some of the older "safe" interfaces in core since they now all belong in webauthn-rs. This removes some old comments and also hides some modules. Finally some other functions are moved to more contextually relevant places. 

### DIFF
--- a/webauthn-rs-core/src/core.rs
+++ b/webauthn-rs-core/src/core.rs
@@ -16,7 +16,6 @@
 #![warn(missing_docs)]
 
 use rand::prelude::*;
-use std::collections::BTreeSet;
 use std::convert::TryFrom;
 use std::time::Duration;
 use url::Url;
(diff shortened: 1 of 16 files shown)


webauthn-rs-core/src/core.rs
```diff
@@ -163,21 +126,19 @@ impl WebauthnCore {
     /// persist. It is strongly advised you associate this RegistrationState with the
     /// UserId of the requester.
     #[allow(clippy::too_many_arguments)]
-    pub fn generate_challenge_register_options(
+    pub fn generate_challenge_register(
```
micolous, line 129: Replace options with a struct with a builder pattern. That will let us encode the defaults which existed on the old `generate_challenge_register()` function.

webauthn-rs-core/src/core.rs
```diff
@@ -843,87 +786,37 @@ impl WebauthnCore {
...
-            policy,
-            extensions,
-            allow_backup_eligible_upgrade,
-        )
-    }
+        let policy = if let Some(policy) = policy {
+            policy
+        } else {
+            let mut policies =
+                BTreeSet::from_iter(creds.iter().map(|cred| cred.registration_policy.to_owned()));
```
micolous, line 808: This creates a copy of all `creds` registration policies, and then deduplicates them, and then only actually passes through the first policy.

This could instead just copy the first registration policy it sees, then checks that all subsequent policies are the same. If one differs, break immediately with `InconsistentUserVerificationPolicy`.

If `creds` is empty, then break with `CredentialNotFound`.

micolous (approved):
CI failures look pre-existing (https://github.com/kanidm/webauthn-rs/issues/407), which can be handled here or in a follow-up PR.

PR #410, opened 2024-01-02
Add fake credential generator interfaces

### LINKED ISSUE #362: Randomly generate credentials as an enumeration defence
Some IDP's wish to "hide" their users names during login, so they will present a false dialog prompt that says "username/password" incorrect. This can work with rk's because webauthn sends an empty allowed credential id list. But if a user *does* have credentials that aren't rk, they are listed.

In this case we should be able to "synthesise" random credentials from a seed + username so that it "appears" there are credentials to any/every username, even though they aren't actually true or viable credentials.

Firstyear:
Fixes #362 - adds a fake credential ID generator so that RP's may choose to implement account enumeration defences.

### DIFF
--- a/webauthn-rs-core/src/fake.rs
+++ b/webauthn-rs-core/src/fake.rs
@@ -199,7 +199,7 @@ impl FakeCredentialIDDistribution for FakePasskeyDistribution {
 /// To achieve this, the generator uses a seeded CSPRNG for each operation. The CSPRNG is seeded
 /// from the HMAC of the username. The HMAC is keyed from an input that the site provides.
 ///
-/// The HMAC key should not be disclosed, as knowledge of the HMAC key will allow an external
+/// The HMAC key MUST NOT be disclosed, as knowledge of the HMAC key will allow an external
 /// party to determine which IDs are generated and which are not.
 pub struct WebauthnFakeCredentialGenerator<D>
 where
@@ -214,8 +214,25 @@ impl<D> WebauthnFakeCredentialGenerator<D>
 where
     D: FakeCredentialIDDistribution,
 {
+    /// Generate a new random HMAC key for the credential generator. You MUST persist this key for
+    /// future use.
+    pub fn new_hmac_key() -> Result<Vec<u8>, WebauthnError> {
+        let mut key = vec![0; 16];
+
+        openssl::rand::rand_bytes(&mut key)
+            .map_err(WebauthnError::OpenSSLError)
+            .map(|_| key)
+    }
+}
+
+impl<D> WebauthnFakeCredentialGenerator<D>
+where
+    D: FakeCredentialIDDistribution,
+{
     /// Create a new HMAC keyed fake credential generator. You should associate a distribution type
-    /// using type annotations.
+    /// using type annotations. The `hmac_key` is a secret value. Disclosure of it will allow an
+    /// external party to determine if `CredentialID`s are genuine or faked. Rotation of this key
+    /// may also allow detection of genuine or fake credentials.
     pub fn new(hmac_key: &[u8]) -> Result<Self, WebauthnError> {
         let hmac_key = pkey::PKey::hmac(hmac_key).map_err(WebauthnError::OpenSSLError)?;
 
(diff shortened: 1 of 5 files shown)


yaleman (approved):
The logic and assumptions seem sound - I'd say we could probably add "what balance of types do you want" initializer if folks wanted to skew the numbers but ... later.

Firstyear:
Waiting on @micolous to review as well :)

webauthn-rs-core/src/fake.rs
```diff
@@ -0,0 +1,323 @@
...
+}
+
+impl<D> WebauthnFakeCredentialGenerator<D>
+where
+    D: FakeCredentialIDDistribution,
+{
+    /// Create a new HMAC keyed fake credential generator. You should associate a distribution type
+    /// using type annotations.
+    pub fn new(hmac_key: &[u8]) -> Result<Self, WebauthnError> {
+        let hmac_key = pkey::PKey::hmac(hmac_key).map_err(WebauthnError::OpenSSLError)?;
```
micolous, line 220: There should be an explicit check that `hmac_key` is a "reasonable length". I feel like the 4 bytes used in the test isn't very long either 😅 

[Some versions of OpenSSL accept zero-length keys](https://github.com/openssl/openssl/issues/13089), and it seems like a very easy way to shoot yourself in the foot.

I'm also worried about a case where someone does something like:

```rust
let cred_gen: WebauthnFakeCredentialGenerator<FakePasskeyDistribution> =
            WebauthnFakeCredentialGenerator::new(b"www.example.com")?;
```

But I'm not sure if it's worth doing anything about it.

webauthn-rs-core/src/fake.rs
```diff
@@ -0,0 +1,323 @@
...
+    use crate::proto::Base64UrlSafeData;
+
+    #[test]
+    fn test_fake_credential_generator() {
+        let _ = tracing_subscriber::fmt::try_init();
+
+        let cred_gen: WebauthnFakeCredentialGenerator<FakePasskeyDistribution> =
+            WebauthnFakeCredentialGenerator::new(&[0, 1, 2, 3]).unwrap();
+
+        let cred_a = cred_gen.generate("a".as_bytes()).unwrap();
```
micolous, line 262: `cred_gen.generate(b"a")`

webauthn-rs-core/src/fake.rs
```diff
@@ -0,0 +1,323 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::{FakePasskeyDistribution, WebauthnFakeCredentialGenerator};
+    use crate::proto::Base64UrlSafeData;
+
+    #[test]
+    fn test_fake_credential_generator() {
```
micolous, line 256: This can return `Result<(), WebauthnError>`, that way you don't need `unwrap()` everywhere.

webauthn-rs-core/src/fake.rs
```diff
@@ -0,0 +1,323 @@
...
+                ])
+            ]
+        );
+
+        // Demonstrate that re-keying the generator yields different generated credential results.
+        let alt_cred_gen: WebauthnFakeCredentialGenerator<FakePasskeyDistribution> =
+            WebauthnFakeCredentialGenerator::new(&[3, 2, 1, 0]).unwrap();
+
+        let alt_cred_a = alt_cred_gen.generate("a".as_bytes()).unwrap();
+        assert!(alt_cred_a != cred_a);
```
micolous, line 312: `assert_ne!`

webauthn-rs/src/lib.rs
```diff
@@ -199,6 +199,11 @@ use webauthn_rs_core::WebauthnCore;
 
 use crate::interface::*;
 
+/// Fake `CredentialID` generator. See [WebauthnFakeCredentialGenerator] for more details.
```
micolous, line 202: `See [WebauthnFakeCredentialGenerator](fake::WebauthnFakeCredentialGenerator) for more details.`

webauthn-rs-core/src/fake.rs
```diff
@@ -0,0 +1,323 @@
...
+        let hmac_key = pkey::PKey::hmac(hmac_key).map_err(WebauthnError::OpenSSLError)?;
+
+        Ok(WebauthnFakeCredentialGenerator {
+            hmac_key,
+            distribution: std::marker::PhantomData,
+        })
+    }
+
+    /// Given a username as a byte slice, generate a set of deterministic `CredentialID`s.
+    pub fn generate(&self, username: &[u8]) -> Result<Vec<CredentialID>, WebauthnError> {
```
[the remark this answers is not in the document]
Firstyear: Some systems allow "empty" to be a unique username (for example, ldap maps this to anonymous). So there may be a reason to still have this work.

webauthn-rs-core/src/fake.rs
Firstyear: Unwraps tell us which line failed though.

webauthn-rs-core/src/fake.rs
Firstyear: We could always just have a small fn that returns a new random key.

micolous (approved):

Issue #397, opened 2023-12-04
Google Titan Security Key USB-C/NFC fails some compatibility tests


arthurgleckler:
### I did this

I used the [compatibility test](https://webauthn.firstyear.id.au/compat_test) with my  Google Titan Security Key USB-C/NFC.

### I expected the following

I expected it to pass all the tests.

### What actually happened

Several tests failed.

### Operating System / Version

I'm using Chrome 119.0.6045.199 on Ubuntu 22.04.

### Any other comments
[titan-compatibility-results.json](https://github.com/kanidm/webauthn-rs/files/13542394/titan-compatibility-results.json)

Thank you for webauthn-rs.  I'm hoping to use it in a site I run.

arthurgleckler:
This is model K40T.

Firstyear:
The attestation failure is expected because this is not a FIDO certified key. Just because they claim "FIDO compatible" doesn't mean it's passed FIDO certification. This is why it's not in the "known fido list". 

The compat results here are showing that the key doesn't support userverification - did the key prompt you for a PIN or for biometrics? from what I'm reading, this appears to be a *single* factor only key, meaning that it will not work with our "passkey" library interfaces. This is a limitation of the key - not webauthn-rs sorry. 

Generally we wouldn't advise the titan key, as it's quite a limited device, and there are better options available (yubikeys especially).

arthurgleckler:
Thank you very much for the quick response.  I suspected that this might be the answer, so I ordered a Yubikey 5C last night.  I'm excited to try it with `webauthn-rs`.

By the way, I'm in the middle of writing a command-line wrapper for `webauthn-rs` that works similarly to the Tide tutorial example.  Instead of using a web server, though, it takes its input and produces its output in JSON form.  That way, I can call it from a web server written in another programming language (Scheme, in my case) without having to deal with the complexity of an FFI.  If this is at all interesting to you, I'd be happy to contribute the command-line wrapper to the project if and when I get it to work.

Firstyear:
@micolous What do you think? I think it's an interesting idea, and could be a useful debugging/inspection tool, but my brain is also blanking on when I'd need it as a debugging/inspection tool? 

As the same time, it also seems like a pretty thin layer over webauthn-rs, so maintenance wise it would be "low effort". 

@arthurgleckler As always, happy to help with anything else. Good luck with the yubikeys (I have a fair few myself and I really do support that they are the best keys available!)

micolous:
[...]

PR #414, opened 2024-01-13
webauthn-rs command line

arthurgleckler:
Fixes # This pull request implements a new feature, not a fix.

This pull request implements the basic `webauthn-rs` command line that I mentioned in the discussion of issue #397.  The idea is that this wrapper can be used for debugging and experimentation.  It can also be used to implement `webauthn` in programming languages other than Rust.  This is simplified by passing JSON to and from the command-line process rather than using a foreign function interface.

If you run `make` under `tutorial/server/cli/`, you'll see the result of running the command line for the two steps of passkey registration and the two steps of authentication.  It's modeled after the `tutorial/server/tide/` example.

Using this wrapper, I've implemented, in some code that is still private, a webauthn registration and login system in the programming language Scheme, with a Javascript front end.  I've tested it with both a Yubikey 5C key and a Pixel phone passkey.

I've included minimal documentation in the comments, but I'm happy to add more if you think this pull request is a good idea.

Thank you very much for making `webauthn-rs` available.  Without it, implementing `webauthn` for a Scheme front end would have been vastly more work.


Cargo.toml
```diff
@@ -20,6 +20,7 @@ members = [
     # Authenticator CLI,
     "authenticator-cli",
     # Tutorial / Example sites.
+    "tutorial/server/cli",
```
Firstyear, line 23: Given you intend to use this as a server process, it probably shouldn't be in the tutorial :)

Firstyear: Perhaps a [bin] in webauthn-rs itself?

Firstyear (changes requested):
@micolous what do you think?

Cargo.toml
Firstyear: authenticator-cli is for interacting with CTAP2 devices specifically, so I think that's not right here.

I think that webauthn-server-proxy-bin or something else? What's a name that signals that a server calls out to this for processing those operations?

arthurgleckler: So move `tutorial/server/cli/` to `webauthn-server-proxy-bin/`?

Should I also change `name = "webauthn-cli"` to `name = "webauthn-server-proxy"` in what's now `tutorial/server/Cargo.toml`?

Sorry if I'm slow to understand.

Firstyear: Lets decide on the name first before we move it - but once we choose a good name, then "yes" to both of those steps :)

arthurgleckler: I like `webauthn-server-proxy`.  I'll wait for your decision.

Firstyear: Had a chat externally to micolous, we think `webauthn-rp-proxy` is the right name here if that's okay?

arthurgleckler: That's a good name.  Done.

Firstyear: Thanks

Firstyear (dismissed):

Firstyear:
@micolous If you have some item to review as well, that'd be great.

arthurgleckler:
Okay, I've run rust fmt.  Sorry, I'm a novice Rustacean.

Firstyear:
Its okay, that's what the checks are for. You've done pretty well for a first time :)

arthurgleckler:
Please let me know if there's anything else I should do.  Thanks.

Firstyear:
> Please let me know if there's anything else I should do. Thanks.

@micolous has just been really busy is all :) I'll ask him again to have a look.

arthurgleckler:
Just pinging again.  I'd love to write about this on my blog.  I'm sure that there are other people who could make use of this.  Perhaps I'm flattering myself, but I think a JSON-based API like this with your excellent underlying Rust implementation could allow other programming languages to implement Webauthn more quickly.  Webauthn adoption isn't moving as quickly as it should.

Thanks, and sorry if I'm being a pest.  I know that everyone's busy.

Firstyear:
Oh I'm so sorry this fell by the wayside!

webauthn-rp-proxy/src/main.rs
```diff
@@ -0,0 +1,202 @@
...
+}
+
+fn main() {
+    let arguments = App::new("webauthn-rp-proxy")
+	.arg(
+	    Arg::with_name("step")
+		.long("--step")
+		.takes_value(true)
+		.required(true)
+		.possible_values(&["authenticate-start",
```
micolous, line 94: I would use `clap`'s derive macro with `Subcommand` here, eg:

```rust
#[derive(Debug, Subcommand)]
pub enum Step {
    AuthenticateStart,
    AuthenticateEnd,
    // TODO
}

#[derive(Debug, clap::Parser)]
#[clap(about = "TODO")]
pub struct CliParser {
    #[clap(subcommand)]
    pub step: Step,
    // add other global args here, eg:
    #[clap(long)]
    pub pretty_print: bool,
}
```

Then using it becomes much simpler:

```rust
let opt = CliParser::parse();
let r = match opt.step {
    Step::AuthenticateStart => authenticate_start(&buffer).unwrap();
    // add other steps
};
```

Everything is checked at compile time, so no need to have string constants or `unreachable!()`.

You should also check the result from those functions (ie: `unwrap()` or `expect()`), because they're passing back errors from serde's serialisation library, and then you've ignored them (with `let _ = ...`). While that silences clippy, but doesn't actually fix the issue. 😄

webauthn-rp-proxy/src/main.rs
```diff
@@ -0,0 +1,202 @@
...
+// languages by doing simple JSON I/O to this program.
+
+#[derive(Deserialize)]
+struct RegisterStartRequest {
+    exclude_credentials: Vec<CredentialID>,
+    rp_id: String,
+    rp_origin: String,
+    user_display_name: String,
+    user_name: String,
+    uuid: String,
```
micolous, line 39: You can use a `Uuid` type here if you add the `serde` feature to the `uuid` crate.

webauthn-rp-proxy/Makefile
micolous, line 1: These should be part of a cargo test target, and check the tool's output is valid JSON.

You can do that with a binary target with `assert_cmd`:

* https://docs.rs/assert_cmd/latest/assert_cmd/
* https://rust-cli.github.io/book/tutorial/testing.html#testing-cli-applications-by-running-them

It'd also be good to have some negative assertions – things that are invalid that should be rejected.

webauthn-rp-proxy/src/main.rs
```diff
@@ -0,0 +1,202 @@
+use clap::{App, Arg};
+use serde::{Deserialize, Serialize};
+use serde_json::Result;
+use std::io::{self, Read};
+use uuid::Uuid;
+use webauthn_rs::prelude::*;
+
+// A simple command-line wrapper around webauthn-rs
```
micolous, line 8: If you make these rustdoc module comments (lines starting with `//!`), then this information will show up in the built documentation (`cargo doc`) and rendered in markdown.

webauthn-rp-proxy/src/main.rs
```diff
@@ -0,0 +1,202 @@
...
+
+// - register-start
+// - register-finish
+// - authenticate-start
+// - authenticate-finish
+
+// The program's --step argument specifies which step from the list
+// above is being invoked.  It reads the corresponding JSON Request
+// struct below from stdin, carries out the webauthn step, and writes
+// the JSON Response struct to stdout if no errors occur.
```
micolous, line 21: The protocol here is strange: the tool will send JSON if everything is fine, but then raw strings if there's an error.

Those errors could be collected up into JSON as well – it's still good to send a non-zero exit code, but it means you can structure the errors a bit better on the other side, as there may be cases where user input (ie: things from the user's authenticator, rather than implementation errors in your app) cause it to be rejected.

Being able to handle that in a structured way on the app's side would be nice.

Essentially you'd serialise a `Result` type, like `Result<RegisterStartResponse, WebauthnError>`. I'm not sure if `serde_json` will serialise a `Result` properly, so you may end up with your own `enum` there instead.

micolous (changes requested):
Sorry for the delays! I think what you've got is a good start – the suggestions here would make things more robust and follow rust-isms better. Thanks!

webauthn-rp-proxy/src/main.rs
arthurgleckler: Thank you very much for the thorough review.  I'm a novice at Rust, and have learned a lot from working through your feedback.

I ended up rebasing and pushing again, and Github no longer displays your review comments in the Files changed view.  Sorry about that.  But at least they're still here, in the Conversation page.

I've switched to Subcommand, as suggested, and have captured errors as JSON, too.

webauthn-rp-proxy/src/main.rs
arthurgleckler: Done.

webauthn-rp-proxy/Makefile
arthurgleckler: Done.   See `integration_test.rs`.  The tests check various fields of the JSON return values, including `client`, `server`, `error`, and some fields deeper inside the result.  They check one case of malformed input, too.

webauthn-rp-proxy/src/main.rs
arthurgleckler: Done.

webauthn-rp-proxy/src/main.rs
arthurgleckler: You're absolutely right.  I was using the fail-fast approach, relying on the exit code to signal failure, but it's better to collect error information and deliver it to the caller.  Errors are now put into a JSON object that includes the single field `error`, which is a string from the underlying `Err` object.

arthurgleckler:
I believe that I've incorporated all of your suggestions.  Would you please take a look when you have a chance?

Thanks.

webauthn-rp-proxy/src/main.rs
micolous: Yep, I've written https://github.com/kanidm/webauthn-rs/pull/426 to fix that up.

You probably want to rebase your PR on that and have a look at the output, because don't think the formatting has come out the way you intended.

arthurgleckler: Fixed.  Thanks.

Sorry for all the extra Git traffic.  I believe I've rebased on top of origin/master now, so the diffs are correct.

arthurgleckler:
Now that a506d89 is committed, I've rebased and force-pushed onto that.  I don't know whether that will fix the PR run failure I received email about.

Firstyear:
I restarted the CI to check :)

arthurgleckler:
It failed again, but the underlying error appears to be "error: package `bumpalo v3.15.4` cannot be built because it requires rustc 1.73.0 or newer, while the currently active rustc version is 1.70.0".  I'm not using bumpalo, at least not directly.  Can you think of any reason this would fail?

Oh, I see that it's failing on master, too:

https://github.com/kanidm/webauthn-rs/commits/master/

webauthn-rp-proxy/src/main.rs
micolous: No worries :)

webauthn-rp-proxy/src/main.rs
micolous: Thanks for sticking with this. 😄

micolous (approved):
Yup, `bumpalo` breakage is not your fault. #428 should fix that.

Otherwise things look good now, thanks for sticking with this. 😄

Firstyear (approved):

arthurgleckler:
> Otherwise things look good now, thanks for sticking with this. 😄

I'm thrilled!  Thanks.

Firstyear:
@arthurgleckler This has now been released on crates.io. Thank you!

PR #424, opened 2024-04-01
Replace `println!` from library functions with configurable logging

mishazharov:
Developers building on this crate do not necessarily need logging to occur over `println!` when authenticating 2FA, so this commit replaces the print macros with logging macros

Fixes #

### DIFF
--- a/webauthn-authenticator-rs/src/mozilla.rs
+++ b/webauthn-authenticator-rs/src/mozilla.rs
@@ -80,7 +80,7 @@ impl MozillaAuthenticator {
                         continue;
                     }
                     PinError::InvalidPin(attempts) => {
-                        debug!(
+                        error!(
                             "Wrong PIN! {}",
                             attempts.map_or("Try again.".to_string(), |a| format!(
                                 "You have {} attempts left.",
@@ -96,9 +96,7 @@ impl MozillaAuthenticator {
                         error!("Too many failed attempts in one row. Your device has been temporarily blocked. Please unplug it and plug in again.")
                     }
                     PinError::PinBlocked => {
-                        error!(
-                            "Too many failed attempts. Your device has been blocked. Reset it."
-                        )
+                        error!("Too many failed attempts. Your device has been blocked. Reset it.")
                     }
                     e => {
                         error!("Unexpected error: {:?}", e)
--- a/webauthn-authenticator-rs/src/mozilla.rs
+++ b/webauthn-authenticator-rs/src/mozilla.rs
@@ -67,7 +67,7 @@ impl MozillaAuthenticator {
                     debug!("STATUS: success using device: {}", dev_info);
                 }
                 Ok(StatusUpdate::SelectDeviceNotice) => {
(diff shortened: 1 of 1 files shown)


webauthn-authenticator-rs/src/mozilla.rs
```diff
@@ -80,7 +80,7 @@ impl MozillaAuthenticator {
                         continue;
                     }
                     PinError::InvalidPin(attempts) => {
-                        println!(
+                        debug!(
```
yaleman, line 83: this should be an error

yaleman (changes requested):

webauthn-authenticator-rs/src/mozilla.rs
mishazharov: Thanks for taking a look, updated!

mishazharov:
It looks like there's some CI failures unrelated to this change:
```
error: package `bumpalo v3.15.4` cannot be built because it requires rustc 1.73.0 or newer, while the currently active rustc version is 1.70.0
Either upgrade to rustc 1.73.0 or newer, or use
cargo update -p bumpalo@3.15.4 --precise ver
where `ver` is the latest version of `bumpalo` supporting rustc 1.70.0
Error: Process completed with exit code 101.
```

yaleman:
@Firstyear how's this one look? Figured I'd ask because you've used it more

yaleman (dismissed):

webauthn-authenticator-rs/src/mozilla.rs
```diff
@@ -58,19 +58,19 @@ impl MozillaAuthenticator {
...
-                    println!("STATUS: device unavailable: {}", dev_info)
+                    debug!("STATUS: device unavailable: {}", dev_info)
                 }
                 Ok(StatusUpdate::Success { dev_info }) => {
-                    println!("STATUS: success using device: {}", dev_info);
+                    debug!("STATUS: success using device: {}", dev_info);
                 }
                 Ok(StatusUpdate::SelectDeviceNotice) => {
-                    println!("STATUS: Please select a device by touching one of them.");
+                    debug!("STATUS: Please select a device by touching one of them.");
```
Firstyear, line 70: I think this one needs to be info! at the least.

Firstyear (commented):
Actually @micolous would be the person to ask. There are some places where I think we use these messages as communication points for cli tools.

micolous:
> Actually @micolous would be the person to ask. There are some places where I think we use these messages as communication points for cli tools.

All the other backends use `info!`, `error!`, etc.; the Mozilla backend is the only one which *doesn't*.

The CLI tools I wrote touch the other backends (as they need to interact with it on a CTAP2 level, rather than an "Authenticator" level), so this would at least make things consistent (as they will eventually replace the Mozilla backend).

I believe the main user of the Mozilla backend is Kanidm's CLI tools – I don't know what makes sense there.

yaleman:
Libraries really should use the `log` primitives instead of `println` so the dependent apps can do what they need to from there...

Firstyear (approved):

PR #435, opened 2024-04-26
Add support for Webauthn L3 options

Firstyear:


webauthn-rs-proto/src/options.rs
```diff
@@ -179,6 +179,56 @@ impl ToString for AuthenticatorTransport {
...
 
+/// The type of attestation on the credential
+///
+/// <https://www.iana.org/assignments/webauthn/webauthn.xhtml>
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Hash)]
+pub enum AttestationFormat {
+    /// Packed attestation
+    #[serde(rename = "packed", alias = "Packed")]
+    Packed,
+    /// TPM attestation (like Micrsoft)
```
micolous, line 190: Microsoft

webauthn-rs-proto/src/options.rs
```diff
@@ -179,6 +179,56 @@ impl ToString for AuthenticatorTransport {
...
+    #[serde(rename = "tpm", alias = "Tpm", alias = "TPM")]
+    Tpm,
+    /// Android hardware attestation
+    #[serde(rename = "android-key", alias = "AndroidKey")]
+    AndroidKey,
+    /// Older Android Safety Net
+    #[serde(
+        rename = "android-safetynet",
+        alias = "AndroidSafetyNet",
+        alias = "AndroidSafteynet"
```
micolous, line 200: typo ("saftey")?

webauthn-rs-proto/src/options.rs
```diff
@@ -179,6 +179,56 @@ impl ToString for AuthenticatorTransport {
...
+    )]
+    AndroidSafetynet,
```
micolous, line 202: This changed from `AndroidSafetyNet` to `AndroidSafetynet` in the move.

It looks like `SafetyNet` is the usual capitalisation:

* https://developer.android.com/privacy-and-security/safetynet
* https://developer.android.com/privacy-and-security/safetynet/attestation
* https://developer.android.com/privacy-and-security/safetynet/deprecation-timeline

webauthn-rs-core/src/core.rs
```diff
@@ -892,6 +919,7 @@ impl WebauthnCore {
         policy: Option<UserVerificationPolicy>,
         extensions: Option<RequestAuthenticationExtensions>,
         allow_backup_eligible_upgrade: Option<bool>,
+        hints: Option<Vec<PublicKeyCredentialHints>>,
```
micolous, line 922: I feel like with a _4th_ optional parameter, now is a good time to move this function to a builder pattern, like `ChallengeRegisterBuilder` does.

I'm not sure whether this is a public API, but `WebauthCore::generate_challenge_authenticator` could then become a wrapper for the new builder pattern, and marked as deprecated.

webauthn-rs/src/lib.rs
```diff
@@ -1161,6 +1169,17 @@ impl Webauthn {
             .user_verification_policy(UserVerificationPolicy::Required)
             .reject_synchronised_authenticators(true)
             .exclude_credentials(exclude_credentials)
+            .hints(Some(
+                // hybrid can NOT perform attestation
```
micolous, line 1173: **nit:** `s/can/do/` (also below)

I believe they technically *could*, I can share a USB FIDO2 device over caBLE. Just Apple and Google's implementations (with synchronised credentials stored in iCloud Keychain / Google Password Manager) don't.

Also, the documentation of this function says that it "precludes the use of (synchronised) authenticators"; this should be reworded to be more specific. Something like:

----

Hardware attestation requires that private key material is bound to a single hardware authenticator, and cannot be copied or moved out of it. At present, *all* widely deployed `Hybrid` authenticators (Apple iCloud Keychain and Google Passkeys in Google Password Manager) are synchronised authenticators which can roam between multiple devices, and so *can never be attested*.

**As of `webauthn-rs` vX.Y**, this creates a challenge with "hints" `(link to PublicKeyCredentialHints)` to only use `ClientDevice` or `SecurityKey` devices, so a user-agent supporting Webauthn L3 won't offer to use `Hybrid` credentials. On user-agents not supporting Webauthn L3, and on older versions of `webauthn-rs`, user-agents would show a QR code and a user could attempt to register a `Hybrid` authenticator, but it would always fail at the end -- which is a frustrating user experience!

webauthn-rs-proto/src/options.rs
Firstyear: Yeah, I renamed it to serilise to android-safetynet, but given I added the manual rename we don't need it.

webauthn-rs-core/src/core.rs
Firstyear: Yeah, probably. I'll do that now too.

webauthn-rs-core/src/core.rs
```diff
@@ -878,21 +938,15 @@ impl WebauthnCore {
...
-    /// credentials being used in the authentication.
-    ///
-    /// `allow_backup_eligible_upgrade` allows rejecting credentials whos backup eligibility
-    /// has changed between registration and authentication.
-    pub fn generate_challenge_authenticate(
+    /// Generate a new challenge builder for client authentication. This is the first
+    /// step in authentication of a credential. This function will return an
+    /// authentication builder allowing you to customise the parameters that will be
+    /// sent to the client.
+    pub fn new_challenge_authenticate_builder(
```
micolous, line 945: Can this be `ChallengeAuthenticateBuilder::new()` instead? It doesn't look like it uses any fields in `WebauthnCore`, so it feels weird to put a constructor here.

micolous, line 945: Empty `creds` = resident keys should be documented too :)

webauthn-rs-core/src/core.rs
```diff
@@ -910,8 +964,37 @@ impl WebauthnCore {
...
+        })
+    }
+
+    /// Generate a new challenge for client authentication from the parameters defined by the
+    /// [ChallengeAuthenticateBuilder].
+    ///
+    /// This function will return the
+    /// [RequestChallengeResponse] which is suitable for serde json serialisation
+    /// to be sent to the client. The client (generally a web browser) will pass this JSON
+    /// structure to the `navigator.credentials.create()` javascript function for registration.
```
micolous, line 982: I'd reword the return part:

---

This function will return:

* a [RequestChallengeResponse], which is sent to the client (and can be serialised as JSON).
 
  A web application would then pass the structure to [the browser's `navigator.credentials.create()` API](https://w3c.github.io/webappsec-credential-management/#dom-credentialscontainer-get) to trigger authentication.

* an [AuthenticationState], which *must* be persisted on the server side. Your application should associate the state with a private session ID, to prevent use in other sessions.

webauthn-rs-proto/src/auth.rs
```diff
@@ -25,6 +25,11 @@ pub struct PublicKeyCredentialRequestOptions {
     pub allow_credentials: Vec<AllowCredentials>,
     /// The verification policy the browser will request.
     pub user_verification: UserVerificationPolicy,
+
+    /// Hints defining which credentials may be used in this operation.
```
micolous, line 29: "which types of credentials"

micolous (dismissed):

webauthn-rs-core/src/core.rs
Firstyear: Yeah, but for consistency we already do the same with ChallengeRegisterBuilder.

micolous: OK

micolous (approved):

PR #437, opened 2024-04-28
Things that you only find during a release

Firstyear:
As is tradition, during a release you find everything that doesn't work like git dependencies, missing versions in the workspace, missing items in Cargo.toml etc. This is a smattering of smaller tidy ups to fix these during the cargo publish process. 

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ members = [
     # Tutorial / Example sites.
     "tutorial/server/tide",
     "tutorial/server/axum",
-    "tutorial/server/actix_web",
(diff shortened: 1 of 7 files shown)


yaleman (approved):
Probably need to make a "no git/paths in the cargo.toml" check somewhere in the release planning 😄

Firstyear:
> Probably need to make a "no git/paths in the cargo.toml" check somewhere in the release planning 😄

We hit a lot of problems with libraries not doing releases with features we needed, but I think mostly they are fixed now?

PR #458, opened 2024-12-11
Upgrade hyper/http, use new APIs

micolous:
**Note:** it compiles, but I need to test with physical hardware.

Upgrades `http`, `http-body-util` and `hyper` to current versions.

`hyper` switched from using Tokio traits to their own custom traits [a while ago](https://github.com/hyperium/hyper/issues/3110), but have an adapter to Tokio traits (`hyper-util`). So this switches everything to use the adapter.

Related to #444 

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -76,10 +76,13 @@ compact_jwt = "0.4.2"
 futures = "^0.3.25"
 hex = "0.4.3"
 http = "^0.2.9"
-http-body = "=1.0.0-rc.2"
-http-body-util = "=0.1.0-rc.2"
-hyper = { version = "=1.0.0-rc.3", default-features = false, features = [
+http-body = "1.0.1"
+http-body-util = "0.1.2"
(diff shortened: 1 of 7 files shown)


Firstyear:
@micolous Given how much this PR is blocking I think we should merge it and commit to testing later.

micolous:
I can't actually find all the hardware pieces I need to test this properly. 😞 

`cable-tunnel-server-*` is mostly a proof-of-concept demo rather than anything intended for serious usage, so I'm less worried if something breaks.

CI was failing because `fido-hid-rs` has a transitive dependency on `home`, which as part of `cargo`, has an MSRV-3 policy (=1.81), which is higher than our MSRV (=1.79).

https://github.com/kanidm/webauthn-rs/pull/461 bumps the MSRV to 1.81. I've included that patch in this PR as well to make the build pass.

yaleman (approved):
Seems legit to me, dependency has been merged as noted

PR #470, opened 2025-01-02
fix axum tutorial in makefile

### LINKED ISSUE #469: Axum tutorial is broken
The axum tutorial seems wildly broken :-(

### I did this
```shell
git clone https://github.com/kanidm/webauthn-rs.git
# optional (same problem with or without): git checkout v0.5.1
cd webauthn-rs/tutorial
make axum
```
### I expected the following
to successfully build the tutorial AND for the tutorial to work (see "Any other comments")

### What actually happened
```
cd server/axum && ./build_wasm.sh && cargo run
Error: `cargo metadata` exited with an error: error: failed to parse manifest at `/Users/dave/developer/rust/webauthn-rs/tutorial/wasm/Cargo.toml`

Caused by:
  error inheriting `serde_json` from workspace root manifest's `workspace.dependencies.serde_json`

Caused by:
  failed to find a workspace root

Caused by: `cargo metadata` exited with an error: error: failed to parse manifest at `/Users/dave/developer/rust/webauthn-rs/tutorial/wasm/Cargo.toml`

Caused by:
  error inheriting `serde_json` from workspace root manifest's `workspace.dependencies.serde_json`

Caused by:
  failed to find a workspace root
```

### Version (and git commit)
[96ddb68](https://github.com/kanidm/webauthn-rs/commit/96ddb68104e62dc2aa855da070cda3057cc44095) but also with tag `v0.5.1`

Rust:  1.83.0
wasm-pack 0.13.1

macOS 15.2

### Any other comments

The build is just the first problem. There are several other things broken with the tutorial code (note: these problems are with the tutorial code only, while the `webauthn-rs/compat_tester/webauthn-rs-demo` does seem to work (at least I am prompted to register a passkey).

After modifying `webauthn-rs/tutorial/wasm/Cargo.toml` and `webauthn-rs/tutorial/server/axum/Cargo.toml` to get the code to build. The server starts listening on port 8080 but the demo does not function.
- The `assets/wasm/index.html` file expects `./wasm.js` and `./wasm_bg.wasm` however, the files available are `./wasm_tutorial.js` and `./wasm_tutorial_bg.wasm`. After modifying the index.html to the new names, the `start_register` appears to work per the server logs but I'm never prompted to register a passkey. The browser (chrome) inspect console shows a list of unintelligible errors from the wasm code.
- Modifying the build feature for `'javascript'` instead of `'wasm'` is also a problem. Going to `https://localhost:8080/` returns a `404` error because `main.rs` code has the following lines:
```rust
    #[cfg(feature = "javascript")]
    let app = Router::new()
        .merge(app)
        .nest_service("/assets", tower_http::services::ServeDir::new("assets/js"));
      ### ^^^^^^^^^^^^^^^^^^^^^^^^^########
     # "/assets" probably should be "/"
```
After making that change, then `assets/index.html` requires `/assets/auth.js` to be modified to `/auth.js`.  But even then the tutorial itself fails to complete registration with seemingly `start_register` being called no prompt to register a passkey.

Firstyear:
Fixes #469

Remove the wasm build from the axum tutorial, as it defaults to the js display anyway. 

### DIFF
--- a/tutorial/Makefile
+++ b/tutorial/Makefile
@@ -13,7 +13,7 @@ actix: wasm
 
 axum: ## Build and run the axum server
 axum:
-	cd server/axum && ./build_wasm.sh && cargo run
+	cd server/axum && cargo run
 
 
 tide: ## Build and run the tide server


dpasirst:
Please note, this currently only partially fixes #469 - specifically, it fixes the broken build, but the tutorial itself still does not work.

yaleman:
@dpasirst I just tested it and it worked fine on my machine?

yaleman (approved):

dpasirst:
@yaleman - It is possible that I'm doing something wrong. Perhaps to figure this out: would you please clarify how you tested? For example, Are you sure you tested in a 100% clean setup with a git pull to a new location (NOTE: `cargo clean` is insufficient because this uses Makefile and there is no `make clean` defined). I ask because with this change, the wasm code is never built or deployed for axum. Therefore, `webauthn-rs/tutorial/server/axum/assets/wasm` would only contain an index.html and no javascript or wasm. It may be worth checking if you have old artifacts in that location.

If you start from a completely clean setup and do the following:
```shell
git clone https://github.com/kanidm/webauthn-rs.git
cd webauthn-rs/tutorial/
make axum
```
The rust code builds and the webserver starts listening, the `index.html` file is served.
BUT:
The resulting web page only says "Welcome to the WebAuthn Server!" but no form is displayed.

`webauthn-rs/tutorial/server/axum/assets/wasm` only contains the `index.html` and nothing else as `webauthn-rs/tutorial/server/axum/build_wasm.sh` was never run.

yaleman:
You're right, I hadn't taken into account having an already-built version, thanks for calling it out - the default feature for axum was set to `wasm` not javascript, so I've fixed that in #471

dpasirst:
@yaleman - yes, that does get the tutorial working for javascript as long as you go to `http://localhost:8080/assets`

Two observations:
1) it is a bit non-obvious to go to "/assets" as opposed to just "/" - the other tutorials (actix_web and tide are just "/")

2) the crate fixes also got the `webauthn-rs/tutorial/server/axum/build_wasm.sh` working - YAY!  but... the output to `webauthn-rs/tutorial/server/axum/assets/wasm` is named `wasm_tutorial.*` which does not align with the content in index.html; thus:

```javascript
        import init, { run_app } from './wasm.js';     <--- This line
        async function main() {
           await init('./wasm_bg.wasm');     <--- This line
           run_app();
        }
```
modify to:
```javascript
        import init, { run_app } from './wasm_tutorial.js';
        async function main() {
           await init('./wasm_tutorial_bg.wasm');
           run_app();
        }
```

After that change, the wasm tutorial for axum will work too!

Issue #477, opened 2025-03-23
Deserializing JSON representation of credentials.create() as RegisterPublicKeyCredential fails


bjorn3:
### I did this

```js
let options = fetch_options_from_server();
options.publicKey = PublicKeyCredential.parseCreationOptionsFromJSON(options.publicKey);
let credential = await navigator.credentials.create(options);
send_to_server(JSON.stringify(credential));
```

```rust
fn register_handler(credential: &str) {
    let credential: RegisterPublicKeyCredential = serde_json::from_str(credential).unwrap();
    // ...
}
```

### I expected the following

Deserialization succeeds.

### What actually happened

Deserialization fails as the browser serialized the extensions under the `clientExtensionResults` key in accordence with the [`PublicKeyCredential.toJSON` docs](https://developer.mozilla.org/en-US/docs/Web/API/PublicKeyCredential/toJSON), while webauthn-rs expects them to be under the `extensions` key.

### Version (and git commit)

webauthn-rs 0.5.1

### Operating System / Version

Firefox for Linux 137.0b9

### Any other comments

Deserializing the result of `credentials.get()` works fine though.

bjorn3:
Turns out the deserialization failure was because I accidentally serialized the credential twice in a row. Still it wouldn't deserialize the extensions correctly.

Firstyear:
Can you give an example of the json that causes the failure?

bjorn3:
```json
{"authenticatorAttachment":"cross-platform","clientExtensionResults":{"credProps":{"rk":false}},"id":"","rawId":"","response":{"attestationObject":"","clientDataJSON":"","publicKey":"","publicKeyAlgorithm":-7,"transports":["usb"]},"type":"public-key"}
```

is parsed as

```rust
RegisterPublicKeyCredential {
    id: "",
    raw_id: Base64UrlSafeData(
        [],
    ),
    response: AuthenticatorAttestationResponseRaw {
        attestation_object: Base64UrlSafeData(
            [],
        ),
        client_data_json: Base64UrlSafeData(
            [],
        ),
        transports: Some(
            [
                Usb,
            ],
        ),
    },
    type_: "public-key",
    extensions: RegistrationExtensionsClientOutputs {
        appid: None,
        cred_props: None,
        hmac_secret: None,
        cred_protect: None,
        min_pin_length: None,
    },
}
```

rather than

```rust
RegisterPublicKeyCredential {
    id: "",
[...]

PR #478, opened 2025-03-25
Show that we can parse json

Firstyear:
Relates #477 

Quick test to show we can parse cred_props. 

### DIFF
--- a/webauthn-rs-proto/src/tests.rs
+++ b/webauthn-rs-proto/src/tests.rs
@@ -1,0 +1,8 @@
+use crate::RegistrationExtensionsClientOutputs;
+
+#[test]
+fn parse_cred_props() {
+    let input = r#"{"credProps":{"rk":false}}"#;
+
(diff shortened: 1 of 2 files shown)


webauthn-rs-proto/src/tests.rs
```diff
@@ -0,0 +1,8 @@
+use crate::RegistrationExtensionsClientOutputs;
+
+#[test]
+fn parse_cred_props() {
+    let input = r#"{"credProps":{"rk":false}}"#;
+
+    let _parsed_extn: RegistrationExtensionsClientOutputs = serde_json::from_str(input).unwrap();
```
micolous, line 7: To test this change, shouldn't this parse a `RegisterPublicKeyCredential` or `PublicKeyCredential` JSON with the with both names for the `extensions` field?

Firstyear: Yeah, the original hypothesis was something else, that's why the test was like this, I'll add the test.

micolous (approved):

PR #479, opened 2025-03-28
authenticator/usb: detect wink capability

lucab:
This adds USB-HID logic for detecting `CTAPHID_WINK` support.

Ref: https://fidoalliance.org/specs/fido-v2.0-ps-20190130/fido-client-to-authenticator-protocol-v2.0-ps-20190130.html#usb-hid-init

---

- [ ] ~~documentation has been updated with relevant examples (if relevant)~~

### DIFF
--- a/webauthn-authenticator-rs/src/usb/mod.rs
+++ b/webauthn-authenticator-rs/src/usb/mod.rs
@@ -54,6 +54,7 @@ pub struct USBToken {
     cid: u32,
     supports_ctap1: bool,
     supports_ctap2: bool,
+    supports_wink: bool,
     initialised: bool,
 }
 
@@ -69,6 +70,7 @@ impl fmt::Debug for USBToken {
             .field("cid", &self.cid)
             .field("supports_ctap1", &self.supports_ctap1)
             .field("supports_ctap2", &self.supports_ctap2)
+            .field("supports_wink", &self.supports_wink)
             .field("initialised", &self.initialised)
             .finish()
     }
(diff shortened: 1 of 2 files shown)


micolous (approved):
Looks good to me, thanks!

One UI caveat with designing an application which uses `WINK` is that the command only works over USB-HID, and not other transports. Also, while the spec suggests that an authenticator could provide a visual *or* audible notification, in practice most USB-HID authenticators exclusively rely on *visual* notifications.

CTAP 2.1 has an `Selection` method, which requires some user action to confirm which authenticator they want to use, whether that is by pressing a button (BLE, USB) or moving the authenticator in range of an NFC antenna.

While the library docs say this is only available on CTAP 2.1 and later (because that’s what the spec says), the library [has a backwards-compatibility helper](https://github.com/kanidm/webauthn-rs/blob/18c8ff5e5c9affd5726b9fb0779f3a64dad0b1b7/webauthn-authenticator-rs/src/ctap2/ctap20.rs#L501) for CTAP 2.0 and 2.1-PRE which issues a `MakeCredential` request.

That behaviour roughly matches what Windows 10’s WebAuthn API does.

There are some USB-HID authenticators that don’t have buttons and instead authenticate whenever you insert them, but thankfully those only support U2F so aren’t supported by the library anyway. 😄

lucab:
@micolous thanks for the feedback.
I just wanted to get a bit more familiar with the FIDO CTAP spec, and started implementing some basic winking logic on my own before discovering the existence of this library.
If it isn't a problem, I'd like to send a PR to implement that as a `fido-key-manager wink` subcommand in the next few days.

Issue #480, opened 2025-03-28
Remove unmaintained 'paste' dependency from base64urlsafedata


ptdecker:
### Short description of issue

base64urlsafedata 0.5.1 is dependent upon paste 1.0.15 which is no longer maintained and flagged by `cargo audit` as such. 

Advisory: (RUSTSEC-2024-0436)[https://rustsec.org/advisories/RUSTSEC-2024-0436]

### Package version (and git commit)

base64urlsafedata 0.5.1

### Operating System / Version

Darwin Todds-MacBook-Pro.local 24.3.0 Darwin Kernel Version 24.3.0: Thu Jan  2 20:24:16 PST 2025; root:xnu-11215.81.4~3/RELEASE_ARM64_T6000 arm64

PR #481, opened 2025-04-02
Bugs

yaleman:
- Fixes some compile warnings that've been coming up for a while.
- Fixes: #480
- Fixes: authenticator-cli was crashing on startup due to double-init tracing

### DIFF
--- a/webauthn-authenticator-rs/src/cable/discovery.rs
+++ b/webauthn-authenticator-rs/src/cable/discovery.rs
@@ -52,7 +52,7 @@ impl DerivedValueType {
     }
 
     pub fn derive(&self, ikm: &[u8], salt: &[u8], output: &mut [u8]) -> Result<(), WebauthnCError> {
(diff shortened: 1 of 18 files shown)


webauthn-authenticator-rs/src/cable/discovery.rs
```diff
@@ -2,7 +2,6 @@
 #[cfg(doc)]
 use crate::stubs::*;
 
-use num_traits::ToPrimitive;
```
micolous, line 5: If you update `num-derive` from `0.3` to `0.4`, all of the non-local `impl` warnings go away. :)

micolous (changes requested):

webauthn-authenticator-rs/src/cable/discovery.rs
yaleman: ... ugh, seriously? I tried updating things and all the conditional compilation shit must have hidden this. 😢

PR #484, opened 2025-04-08
authenticator/usb: add support for wink function

lucab:
This adds support for `CTAPHID_WINK`, i.e. winking function for USB tokens.

Ref: https://fidoalliance.org/specs/fido-v2.0-ps-20190130/fido-client-to-authenticator-protocol-v2.0-ps-20190130.html#usb-hid-wink

- [ ] ~~documentation has been updated with relevant examples (if relevant)~~

Manually tested on a Titan USB-C key:

https://github.com/user-attachments/assets/b25682c2-33f3-48a5-9341-dcff9a82ffab

### DIFF
--- a/fido-key-manager/src/main.rs
+++ b/fido-key-manager/src/main.rs
@@ -209,6 +209,9 @@ pub enum Opt {
     SoloKeyRandom,
     #[cfg(feature = "yubikey")]
     YubikeyGetConfig,
+    #[cfg(feature = "usb")]
+    /// Wink a connected USB device.
+    Wink,
 }
 
 #[derive(Debug, clap::Parser)]
@@ -773,6 +776,21 @@ async fn main() {
 
             println!("YubiKey config:");
             println!("{cfg}")
+        }
+
+        #[cfg(feature = "usb")]
+        Opt::Wink => {
+            use webauthn_authenticator_rs::usb::USBTransport;
+
+            println!("Insert a USB device...");
+            let transport = USBTransport::new().await.unwrap();
+            let mut stream = transport.watch().await.unwrap();
+            while let Some(event) = stream.next().await {
+                if let TokenEvent::Added(mut token) = event {
+                    token.wink().await.expect("failed to wink USB device");
+                    return;
+                }
+            }
         }
     }
 }
--- a/webauthn-authenticator-rs/src/usb/mod.rs
+++ b/webauthn-authenticator-rs/src/usb/mod.rs
@@ -24,7 +24,9 @@ use fido_hid_rs::{
 };
 
 use crate::error::WebauthnCError;
-use crate::transport::types::{KeepAliveStatus, Response, U2FHID_CANCEL, U2FHID_CBOR, U2FHID_INIT};
+use crate::transport::types::{
+    KeepAliveStatus, Response, U2FHID_CANCEL, U2FHID_CBOR, U2FHID_INIT, U2FHID_WINK,
+};
 use crate::transport::*;
 use crate::ui::UiCallback;
 use crate::usb::framing::*;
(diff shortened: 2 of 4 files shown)


fido-key-manager/src/main.rs
```diff
@@ -774,5 +777,20 @@ async fn main() {
...
+        #[cfg(feature = "usb")]
+        Opt::Wink => {
+            use webauthn_authenticator_rs::usb::USBTransport;
+
+            println!("Insert a USB device...");
+            let transport = USBTransport::new().await.unwrap();
+            let mut stream = transport.watch().await.unwrap();
+            while let Some(event) = stream.next().await {
+                if let TokenEvent::Added(mut token) = event {
+                    token.wink().await.expect("failed to wink USB device");
```
yaleman, line 790: Instead of an expect, how 'bout we do
```suggestion
                    return token.wink().await.map_err(|err| eprintln!"failed to wink USB device: {:?}", err);
```

webauthn-authenticator-rs/src/usb/mod.rs
```diff
@@ -206,6 +208,37 @@ impl USBToken {
...
         Response::try_from(&f)
     }
+
+    /// Wink the device.
+    ///
+    /// This performs a vendor-defined action that provides some visual or audible
+    /// identification of the device.
+    pub async fn wink(&mut self) -> Result<(), WebauthnCError> {
+        if !self.initialised {
+            error!("attempted to transmit to uninitialised token");
```
yaleman, line 218:
```suggestion
            error!("Attempted to transmit to uninitialised token");
```

webauthn-authenticator-rs/src/usb/mod.rs
```diff
@@ -206,6 +208,37 @@ impl USBToken {
...
+            return Err(WebauthnCError::Internal);
+        }
+        if !self.supports_wink {
+            error!("token does not support wink function");
```
yaleman, line 222:
```suggestion
            error!("Token does not support the wink function");
```

yaleman (changes requested):
Minor nits but woo, winks!

Firstyear:
@micolous Can you review this too?

fido-key-manager/src/main.rs
lucab: This doesn't work, as the containing function (i.e. `main`) does not return a `Result`. The existing code nearby here is using expects/unwraps to exit the CLI with an error code.
I guess it makes sense to rework error-handling for all of `main` in the longer term.

yaleman: Ah, derp, true. Sigh.

fido-key-manager/src/main.rs
```diff
@@ -774,5 +777,20 @@ async fn main() {
...
         }
+
+        #[cfg(feature = "usb")]
+        Opt::Wink => {
+            use webauthn_authenticator_rs::usb::USBTransport;
+
+            println!("Insert a USB device...");
+            let transport = USBTransport::new().await.unwrap();
+            let mut stream = transport.watch().await.unwrap();
+            while let Some(event) = stream.next().await {
```
micolous, line 788: This will just wink forever until it fails, and doesn't tell you when winks succeeded. A bunch of authenticators will normally flash when you plug them in, so it's hard to tell from the software side if it's doing anything. 😉 

I think this should work a bit more like `info`: by default it only winks the first authenticator which supports blinking, but have a flag which makes it wink forever. And maybe tell you explicitly if a connected device *doesn't* support winking.

micolous (commented):
`webauthn-authenticator-rs` side looks good; I just have reservations about the `fido-key-manager` UX.

fido-key-manager/src/main.rs
lucab: Ack, makes sense, thanks for the feedback.
I've pushed an additional commit on top which I think takes care of all your suggestions.

lucab:
@micolous I've updated the CLI logic, can you have another look at it?

lucab:
@yaleman @Firstyear @micolous I've had this PR open since some time, and I think it doesn't require further changes from my side. Would it be possible to merge it now?

micolous (approved):

PR #504, opened 2025-07-05
6.0-dev - remove openssl

### LINKED ISSUE #499: Drop openssl
We are in the process of dropping OpenSSL wholesale from all Kanidm related projects.

To achieve this in webauthn-rs will be a lot of work. 

The problem is that if we do this as one large PR, this makes it very hard to review.

But if we try to chunk this down to smaller PR's, we either have to keep adding in compat and error handling that we intent to throw out.

I want to suggest we make a branch for staging our changes, so that we can submit openssl related removal changes, in small batches allowing them all to be reviewed. Then at the end, we merge that stage branch as a whole large change to main. This way we can use unwrap/expect during the initially changes, and then remove those as we proceed in the removal.

Firstyear:
**Please refrain from status update requests, user stories and advocacy about OpenSSL, RustCrypto, or alternative solutions (eg: BoringSSL, `aws-lc`, etc.). We've already considered these issues before starting this work, so such commentary only distracts from actually delivering this work.** 🙏  --micolous

## See https://github.com/kanidm/webauthn-rs/issues/499 for current status

Fixes #499 

This is a staging branch, where will merge changes while we prepare to remove OpenSSL


yaleman (dismissed):

aaaronme:
First off, thank you so much for your incredible work on this crate! 
The quality of `webauthn-rs` has been a huge help.

I’m currently working on removing `openssl` from my stack, and this is the final piece of the puzzle. 
Is there a rough timeline or milestone for the v6 release? 
Would love to know when I can make the switch!

Firstyear:
@aaaronme https://github.com/kanidm/webauthn-rs/pull/543 is waiting review then we plan to issue a 6.0-dev release in the coming days. We don't have a final timeline for 6.0 for production yet though as there are some other API changes we wish to make.

PR #509, opened 2025-08-09
fix: elligibility -> eligibility

tessus:
The word `elligibility` is actually spelled with one `l`: `eligibility`

I came across this by reading an error message in another project and found out that it actually came from this code.

This PR fixes the spelling. The only thing I was not quite sure about was changing the error constant, but this crate is still 0.y.z, thus breaking changes are allowed according to semver.

### DIFF
--- a/cable-tunnel-server/backend/src/main.rs
+++ b/cable-tunnel-server/backend/src/main.rs
@@ -162,7 +162,7 @@ impl From<tungstenite::Error> for CableError {
 }
 
 impl CableError {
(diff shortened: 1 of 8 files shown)


Firstyear (approved):

yaleman (approved):
calmed clippy, thanks for the fix!

tessus:
Thanks for taking care of the clippy warnings.

Firstyear (dismissed):

Firstyear (approved):

yaleman (approved):

PR #511, opened 2025-08-13
Change default eligibility

### LINKED ISSUE #510: Migration from CredentialV3 to CredentialV5 seems to be impossible when the Passkey was backed up in the past
### I did this

* Have a `CredentialV3` Passkey that was already backed up and therefore has `backup_state` set to `true`, stored for a user as their 2FA.
* Migrate `CredentialV3` to `CredentialV5` with [Credential::from(other: CredentialV3)](https://docs.rs/webauthn-rs-core/0.5.2/webauthn_rs_core/proto/struct.CredentialV3.html#impl-From%3CCredentialV3%3E-for-CredentialV5)
* Use [start_passkey_authentication()](https://docs.rs/webauthn-rs/latest/webauthn_rs/struct.Webauthn.html#method.start_passkey_authentication) and [finish_passkey_authentication()](https://docs.rs/webauthn-rs/latest/webauthn_rs/struct.Webauthn.html#method.finish_passkey_authentication) to authenticate the user

### I expected the following

When `allow_backup_eligible_upgrade` is allowed `finish_passkey_authentication()` should return an `AuthenticationResult` to make it possible to actually upgrade `backup_eligible`

### What actually happened

`finish_passkey_authentication()` returns the `CredentialMayNotBeHardwareBound` error, which makes it seemingly impossible to migrate those credentials.

### Version (and git commit)

webauthn-rs 0.5.2 (tag 0.5.2)

Linux 6.15.9-zen1-1-zen

### Any other comments

Is this the intended behavior or am I missing something here? Or is there some other way to update the credential before calling `finish_passkey_authentication()`?
Since the `CredentialV3` didn't save the value of `backup_eligible`, it is set to `false` in the conversion to `CredentialV5`. When it could have been `true` for that specific Passkey, when it was setup up originally, which then causes `backup_state` to be true, but the saved `backup_eligible` value to be `false`

Firstyear:
Fixes #510 - default to migrating credentials as backup eligible credentials

### DIFF
--- a/webauthn-rs-core/src/interface.rs
+++ b/webauthn-rs-core/src/interface.rs
@@ -326,14 +326,14 @@ impl From<CredentialV3> for Credential {
             registration_policy,
         } = other;
 
-        // prior to 20220520 no multi-device credentials existed to migrate from.
         Credential {
             cred_id: HumanBinaryData::from(cred_id),
             cred,
             counter,
             transports: None,
             user_verified: verified,
-            backup_eligible: false,
+            // Assume that when we migrate a very old credential that it may have been backup eligible.
+            backup_eligible: true,
             backup_state: false,
             registration_policy,
             extensions: RegisteredExtensions::none(),


yaleman:
Agreed with the above, it's possible to update that yourself and this PR seems like a less-secure default.

Firstyear:
Agreed :)

PR #513, opened 2025-08-16
Allow backup eligibility to change

### LINKED ISSUE #510: Migration from CredentialV3 to CredentialV5 seems to be impossible when the Passkey was backed up in the past

Firstyear:
Fixes #510

### DIFF
--- a/fido-mds/src/lib.rs
+++ b/fido-mds/src/lib.rs
@@ -29,10 +29,10 @@ use crate::mds::AuthenticatorTransport;
 use crate::mds::FidoDevice as RawFidoDevice;
 use crate::mds::FidoMds as RawFidoMds;
 use crate::mds::MetadataStatement as RawMetadataStatement;
+use crate::mds::MultiDeviceCredentialSupport;
 use crate::mds::StatusReport as RawStatusReport;
 use crate::mds::UserVerificationMethod as RawUserVerificationMethod;
 use crate::mds::VerificationMethodAndCombinations;
-use crate::mds::MultiDeviceCredentialSupport;
 use crate::mds::{
     AttestationType, AuthenticationAlgorithm, AuthenticatorGetInfo, BiometricAccuracyDescriptor,
     CodeAccuracyDescriptor, EcdaaAnchor, ExtensionDescriptor, KeyProtection,
--- a/fido-mds/src/mds.rs
+++ b/fido-mds/src/mds.rs
@@ -580,7 +580,6 @@ impl fmt::Display for MultiDeviceCredentialSupport {
     }
 }
 
-
 /// The output of authenticatorGetInfo. Some fields are hidden as they are duplicated
 /// in the metadata statement.
 #[derive(Debug, Clone, Serialize, Deserialize)]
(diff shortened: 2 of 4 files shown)


Firstyear:
That will teach me for submitting a PR in a hurry on a busy day :)

Firstyear:
@BlackDex Yeah, I think I need to just clean this up a bit:

We have all your V3 -> V5 credentials moving to backup_eligible == true.

This means during authentication you have:

* cred.BE = true, auth_data.BE = true (valid)
* cred.BE = true, auth_data.BE = false (currently this FAILS)
* cred.BE = false, auth_data.BE = true (valid, covered by allow_backup_eligible_upgrade)
* cred.BE = false, auth_data.BE = false (valid)

So I think that's the corner case I missed. I'll update the PR to accept this.

Firstyear:
Done

BlackDex:
Sure, I'll see if i can do this today somewhere

BlackDex:
> That sucks :( I see you have merged a separate PR so in this case, do you want me to just close this one since you have a work around?
> 
> I'm so sorry this has been such a pain for you, it is a pretty big mistake on our part.

Ow, no problem. It was fun to figure it out. And i see it more as our own fault for not having updated it more quickly. 

If this could help others or future implementations, then it might be useful to continue this PR. I know Proxmox had a similar issue as Vaultwarden. But there implementation is a bit different then ours.

Also, it shouldn't lower or demote security of course.

Firstyear:
Correct, in this case it doesn't really affect your security stance. Realistically, given how the passkey ecosystem has moved, flags like backup eligibility and such dont matter, as the only way to truly guarantee properties about authenticators is to attest them. If you don't do attestation, the assumption is "anything goes" given how passkey providers operate. 

So I think there is even an argument that we remove this field and flag.

PR #515, opened 2025-08-29
20250829 drop openssl redux

Firstyear:
Another step in dropping OpenSSL - goes through and fixes up the code so all our tests pass again, so that now future changes will be testable! 


Firstyear:
This looks like more changes than it really is due to rebasing with main, so I should probably tidy that.

PR #521, opened 2025-10-16
added PartialEq and Eq to CreationChallengeResponse and its subtypes

hocman2:
This PR implements PartialEq and Eq to CreationChallengeResponse and RequestChallengeResponse along with the subtypes that don't implement it.

It's worth debating wether these objects need to be equality checked at all but it didn't really make sense to me that the outer types didn't implement the traits when most subtypes did, and for the ones that did not, there was no obvious difference as to why the trait was left over for them.
I figured it makes types more uniform regarding their capabilities without compromising security. 

As for actual use, the codebase i'm working on requires the traits on responses that are sent to the user, hence it might be useful to more people.

### DIFF
--- a/webauthn-rs-proto/src/auth.rs
+++ b/webauthn-rs-proto/src/auth.rs
@@ -11,7 +11,7 @@ use crate::options::*;
 use crate::BASE64_ENGINE;
 
 /// The requested options for the authentication
-#[derive(Debug, Serialize, Clone, Deserialize)]
+#[derive(Debug, Serialize, Clone, Deserialize, PartialEq, Eq)]
 #[serde(rename_all = "camelCase")]
 pub struct PublicKeyCredentialRequestOptions {
     /// The challenge that should be signed by the authenticator.
@@ -37,7 +37,7 @@ pub struct PublicKeyCredentialRequestOptions {
 
 /// Request in residentkey workflows that conditional mediation should be used
 /// in the UI, or not.
-#[derive(Debug, Serialize, Clone, Deserialize)]
+#[derive(Debug, Serialize, Clone, Deserialize, PartialEq, Eq)]
 #[serde(rename_all = "camelCase")]
 pub enum Mediation {
     // /// No mediation is provided - This is represented by "None" on the Option
@@ -59,7 +59,7 @@ pub enum Mediation {
 /// for handling. This is meant to be opaque, that is, you should not need
 /// to inspect or alter the content of the struct - you should serialise it
 /// and transmit it to the client only.
-#[derive(Debug, Serialize, Clone, Deserialize)]
+#[derive(Debug, Serialize, Clone, Deserialize, PartialEq, Eq)]
 #[serde(rename_all = "camelCase")]
 pub struct RequestChallengeResponse {
     /// The options.
--- a/webauthn-rs-proto/src/extensions.rs
+++ b/webauthn-rs-proto/src/extensions.rs
@@ -165,7 +165,7 @@ pub struct HmacGetSecretInput {
 /// Extension option inputs for PublicKeyCredentialRequestOptions
 ///
 /// Implements \[AuthenticatorExtensionsClientInputs\] from the spec
-#[derive(Debug, Serialize, Clone, Deserialize)]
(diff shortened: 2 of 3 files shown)


Firstyear:
There isn't a use case for equality of these challenge requests or responses - these are one time structures, that are uniquely returned by the browser/authenticator, and should be processed exactly once and never stored. Nor should they ever be *compared*. 

The correct behaviour if you want to implement these traits on these objects is that they always return `false` on any equality comparison.

hocman2:
First and foremost, just to be on the same page, in the context of Webauthn the relying party is both the server and the client script that initiates authentication flow with the authenticator.

Upon re-reading my first post I realised one thing could be misunderstood, indeed the challenge must be generated and stored server side, i'm not questioning that. But the rest of the payload could very well be formed client side as this is not a violation of the standard at all. Besides, if a client generated their own challenge any spec compliant RP server would have to refuse the authentication/creation attempt anyway.

Now I think we are stepping a bit far from the initial issue with the aforementioned so let's focus back on the initial topic.
Why do you claim that using equality on two `PublicKeyCredentialCreationOptions` implies a security flaw ? Comparing two payloads does not mean you are re-using a challenge.
There are many imaginable cases where it's a nice convenience to have PartialEq/Eq be implemented on these types, some are more valid than others but I don't think a library should be judgmental, either through it's design or it's philosophy, of the user's codebase.

My point is, from the scope of the crate, you can't make that much assumption as to what a valid RP server should do so there is no point in being overly restrictive.

> If these structures ever compared to true, then you have a security vulnerability in your application.

Uh ? That remains to be demonstrated as I don't see a direct correlation.
I specified the only two security measures to take when delivering these payloads and equality check does not create any of these vulnerabilities. 
Having unique challenges is not a security measure it's a strict spec requirement.

Now I'd rather see this not merged than having a straight up logically broken trait implementation that always returns false. This would be non-sensical.

Firstyear:
> 
> Now I think we are stepping a bit far from the initial issue with the aforementioned so let's focus back on the initial topic. Why do you claim that using equality on two `PublicKeyCredentialCreationOptions` implies a security flaw ? Comparing two payloads does not mean you are re-using a challenge. 

The challenge is in the PKCCO. If you compare them and have well formed, unique challenges the comparison is always FALSE. 

And if it ever comes back TRUE, then congrats, you just confirmed you have a CVE in your application allowing authentication replay attacks to occur. 
 


Again - why do you need to do comparison *at all*. That's the underlying issue here. You are trying to achieve something that I don't understand, and you are presenting this need to do comparison, but WHY.

Firstyear:
> Having unique challenges is not a security measure it's a strict spec requirement.

I think this is the point you may be misunderstanding - it's a strict spec requirement because it's THE most critical part of security in webauthn.

Every other parameter that the authenticator signs is deterministic - only the *challenge* changes. This naturally changes the output of the signature, and how the server validates that signature. 

If you ever reuse a challenge, then an attacker can replay older, previously signed responses *without access to the original authenticator*. This allows phishing and mitm attacks. 

That's why it's so critical that these are always unique, because without this, you have no proof that the authenticator is currently in use. Failure to ensure unique challenges is a CVE.

hocman2:
I perfectly understand the need and the purpose of webauthn unique challenges, I never questioned them.
PartialEq just implies you are just comparing two payloads, that's it. 123 == 123 -> true; 123 == 124 -> false
You can't deduce that the application is vulnerable to replay attacks because of that, there is too little information at hand.
Any further assumption must be demonstrated.

As to why anyone could need this, the short answer is: convenience
The PartialEq/Eq traits are trivial operations which many codebases expect. You could say some of them are not well designed but thats judgmental, and as I said earlier, I don't think a library should be judgmental. 
These codebases could be huge and a simple change of API to get rid of Eq necessity could take weeks.
Maybe the response payload is part of a broader response which itself requires PartialEq at some point, this implies the trait has to be manually implemented because webauthn_rs payload's doesn't have it for some obscure reason. That was my first exemple.
There is a realm of possibilities I can't even think of where these traits could be needed, maybe to do the dumbest things, sure, but also maybe because the application has specific constraints where having them would cut down the monkey code.

PR #526, opened 2025-11-22
Update packages

yaleman:
Fixes a bunch of old stuff and some deprecations in things like rand - CI was failing due to a package "home" being deprecated/requiring newer rust.


Firstyear:
Only concern I have is we need a similar PR in the 6.0 branch at some point.

micolous:
Yeah, there are a lot of other changes in the 6.0 branch (which also affect the dependency tree), so I'd keep such a change to that branch.

PR #533, opened 2026-02-06
20260206 6.0 remove openssl tide

Firstyear:
Everything except webauthn-authenticator-rs is now openssl free (#499).

The MAJOR caveat is that we have to remove support for ed25519, ed448 and p521. 

However, *all three* were not in our default list of advertised cryptographic algorithms. This should limit the scope of any damage. 

I think we should still maintain 5.0 on a maintenance branch though for some time after we make main 6.0 

So this merges to the 6.0 dev branch and tomorrow I'll look at the cable and mozilla side of things. Once those are done, then I think we move 5.x to a maintenance branch and we merge this to main and work from there. 


Firstyear:
Not sure where all these conflicts are from, I did a merge before ....

Firstyear:
Fixed :)

micolous (approved):

yaleman (changes requested):
Minor nit but otherwise, byyyyeeeeeopenssl

PR #534, opened 2026-02-07
6.0: Remove caBLE OpenSSL dependency

micolous:
* Migrate `cable` from OpenSSL to RustCrypto (#499)
* Add `crypto_openssl` dependency to `softtoken` (temporarily)
* Remove all OpenSSL `crypto` functions, which were only used by the `cable` feature
* Update crate docs about where OpenSSL is required
* Update some outdated caBLE docs
* Fix many clippy lints

CI changes:

* Don't install OpenSSL on Windows except when `softtoken` is enabled (which still needs OpenSSL), so we can verify we are really not using OpenSSL anymore.
* Make `clippy` check any feature combination on Windows if `win10` is an enabled feature (ie: we haven't checked it on Linux), because OpenSSL isn't required for any of the other feature sets.

Requires:

* [x] #539
* [x] #540 
* [x] https://github.com/kanidm/crypto-glue/pull/22

## Progress

- [x] replace OpenSSL in `crypto`:
  - [x] `regenerate` (removed)
  - [x] `ecdh`
  - [x] `get_group` (removed)
  - [x] `public_key_from_bytes`
  - [x] `point_to_bytes` (removed, but replacement usages need https://github.com/kanidm/crypto-glue/pull/22)
  - [x] `public_key_from_private` (removed)
- [x] replace OpenSSL in `cable::discovery`:
  - [x] `EcKey`
  - [x] `MessageDigest`
  - [x] `pkey`
  - [x] `rand_bytes`
  - [x] `Signer`
- [x] replace OpenSSL in `cable::noise`:
  - [x] `EcKey`
  - [x] `EcPoint`
  - [x] `pkey`
  - [x] `decrypt_aead`
  - [x] `encrypt_aead`
- [x] replace OpenSSL in `cable::tunnel`:
  - [x] `EcKeyRef`
  - [x] `pkey`
- [x] Remove OpenSSL from `webauthn-authenticator-rs` CI jobs on Windows
- [x] Cherry-pick other PRs into this branch:
  - [x] #537
  - [x] #538
- [x] Get this working with actual hardware
  - [x] Make test failures with actual hardware reproducible with CI
  - [x] Test `webauthn-rs` initiator with:
    - [x] Apple authenticator
    - [x] Android authenticator
  - [x] Test `webauthn-rs` authenticator with:
    - [x] Apple initiator
    - [x] Android initiator
    - [x] Windows initiator

## ...

- [x] cargo test has been run and passes
- [x] documentation has been updated with relevant examples (if relevant)

PR #536, opened 2026-02-11
cable_tunnel: Port to RustCrypto (#499)

micolous:
Replace `openssl::rand::rand_bytes` with `crypto_glue::rand` (#499).

This currently won't build due to existing breakages in the `6.0-dev-drop-openssl` branch when the `cable` feature is enabled. I'm looking at those in #534.


micolous:
Closing this in favour of #534

PR #538, opened 2026-02-12
Add support for advertising with Bluez

micolous:
Follow-on from #537, for testing #534 (so I want it in both branches to compare).

This adds support for acting as a caBLE authenticator with Bluez on Linux to `cable_tunnel`.

Unfortunately, this means we always pull in `bluer` as a `dev-dependency` on Linux (and thus depend on `libdbus-1-dev`), but that's already a dependency when the `bluetooth` or `cable` features are enabled.

Tested with `webauthn_authenticator_rs`, iOS 26 and Android as an initiator. Android registered OK, but only offered to authenticate with Google Password Manager (there was not even a caBLE option, even with non-RK non-UV). The others worked fine.

- [x] Test with an existing caBLE implementation

### DIFF
--- a/webauthn-authenticator-rs/examples/cable_tunnel.rs
+++ b/webauthn-authenticator-rs/examples/cable_tunnel.rs
@@ -1,4 +1,39 @@
-//! `cable_tunnel` shares a [Token] over a caBLE connection.
+//! Share a [`Token`][webauthn_authenticator_rs::transport::Token] to a caBLE initiator.
+//!
+//! This program performs the authenticator (ie: mobile device) role of the session. This can tunnel
+//! to a real (physical) device, or share a SoftToken (with `--features softtoken`).
+//!
+//! This requires a BTLE adapter and platform that can send raw advertisement data, either:
+//!
+//! - a Linux machine running Bluez, or
+//! - [a **serial** BTLE HCI adapter][0]
+//!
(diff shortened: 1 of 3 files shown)


webauthn-authenticator-rs/examples/cable_tunnel.rs
```diff
@@ -1,4 +1,39 @@
...
+//! # Expose it to an initiator via its caBLE URL using Bluez (Linux)
+//! cargo run --example cable_tunnel \
+//!     --features cable,crypto,ui-cli,qrcode,softtoken \
+//!     -- \
+//!     --bluez \
+//!     --softtoken-path /tmp/softtoken.dat \
+//!     --cable-url FIDO:/...
+//! ```
+//!
+//! You can also a cropped screenshot of a caBLE QR code with `--qr-image`, instead of using
```
yaleman:
```suggestion
//! You can also try to parse a cropped screenshot of a caBLE QR code with `--qr-image`, instead of using
```

yaleman (dismissed):
Missing word in the docstring

webauthn-authenticator-rs/examples/cable_tunnel.rs
micolous: Done

Firstyear (approved):

PR #544, opened 2026-04-03
6.0: API change: Make `WebauthnAuthenticator` a trait

micolous:
Change `WebauthnAuthenticator` from a `struct` to a (sealed) `trait`, and update local usage. Using a `trait` makes it easier to work with in `dyn` contexts where an application is built with support for multiple backends.

Because this is a breaking API change, it targets the 6.0 branch, even though it doesn't really touch OpenSSL.

When migrating the `authenticate` example to `WebauthnAuthenticator`, it failed Origin checks. I've updated this to use a custom Origin (`demo.webauthn-authenticator-rs.example`). I've also made it use a timestamped user name. This makes it easier to manage these entries in synchronised credential managers, and identify the source (rather than `localhost`).

I've also updated the docs for `WebauthnAuthenticator` and `AuthenticatorBackend` to describe their intended use cases.

Fixes #

### DIFF
--- a/webauthn-authenticator-rs/examples/authenticate.rs
+++ b/webauthn-authenticator-rs/examples/authenticate.rs
@@ -4,12 +4,12 @@ extern crate tracing;
 #[cfg(feature = "softtoken")]
 use std::fs::OpenOptions;
 use std::io::{stdin, stdout, Write};
-use std::time::Duration;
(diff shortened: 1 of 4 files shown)


Firstyear (approved):
Looks good, the trait approach is much nicer.

PR #512, opened 2025-08-13
Update MDS

Firstyear:
Fixes #508 update MDS again

- [x] cargo test has been run and passes
- [ ] documentation has been updated with relevant examples (if relevant)

PR #546, opened 2026-04-20
Accept 2026-Q1 FIDO MDS schema additions

MasKusuno:
The live FIDO MDS blob served from `https://mds.fidoalliance.org/` on 2026-04-20 contains several fields and enum variants that fail deserialization against the current `deny_unknown_fields` contract in `MetadataStatement` and `AuthenticatorGetInfo`. This PR adds the new shapes, preserving `deny_unknown_fields` and keeping all existing tests green.

### `MetadataStatement`
- `iconDark` — dark-mode variant of `icon`
- `providerLogoLight` — provider-branded logo (light theme)
- `providerLogoDark` — provider-branded logo (dark theme)

### `AuthenticatorGetInfo`
- `longTouchForReset`
- `authenticatorConfigCommands`
- `encCredStoreState`
- `encIdentifier`
- `maxPINLength`
- `pinComplexityPolicy`
- `pinComplexityPolicyURL`
- `transportsForReset` (see note below)
- `uvCountSinceLastPinEntry`

### `AuthenticatorTransport`, `AttachmentHint`
- `smart-card`

### `AuthenticatorVersion`
- `FIDO_2_2`
- `FIDO_2_3`

## Typing choices

Fields that fido-mds has no semantic consumer for are typed as `Option<serde_json::Value>` / `Vec<serde_json::Value>` to avoid speculating on values the spec doesn't yet constrain.

`transportsForReset` in particular is modelled as `Vec<serde_json::Value>` because the live blob uses **both** lowercase (`\"usb\"`, `\"nfc\"`) and uppercase (`\"USB\"`, `\"NFC\"`) spellings in the same field. Rather than introduce a case-insensitive deserializer for a field nothing in fido-mds consumes, I kept it opaque. If you'd prefer a case-insensitive `AuthenticatorTransport` deserializer I'm happy to switch.

Existing enum variants (`AuthenticatorTransport::SmartCard`, `AttachmentHint::SmartCard`, `AuthenticatorVersion::Fido2_2` / `Fido2_3`) gained matching `Display` arms and `FromStr` entries where applicable.

All new `MetadataStatement` / `AuthenticatorGetInfo` fields carry `#[serde(default)]` so older blobs remain parseable.

## Reproduction

Before:
```
$ cargo run --release    # against live mds.fidoalliance.org blob
ERROR fido_mds::mds: serde_err=Error(\"unknown field `friendlyNames`, expected one of ...\")
HIGH err: Serde
```
(and several more, uncovered one-by-one as each is patched: `iconDark`, `providerLogoLight`, `smart-card`, `FIDO_2_2`, `FIDO_2_3`, then the AGI cascade.)

After:
```
$ cargo run --release
blob bytes = 8614519
HIGH ok, fido2=300 u2f=127
RAW ok, entries=478
```

## Testing

- `cargo test -p fido-mds`: all 5 existing tests pass.
- Live blob parses end-to-end through both `FidoMds::from_str` (high) and `RawFidoMds::from_str` (raw).

## Context

This is essentially the same shape as #455 / #512 / #532 — FIDO continues to add fields to the MDS payload ahead of corresponding spec revisions. I understand the upstream stance from PR #460 (\"It would be lovely if FIDO stopped changing their MDS formaT on days ending in 'y'\") and this PR sticks to that pattern: accept the new shapes permissively and keep `deny_unknown_fields` on.

If you'd rather relax `deny_unknown_fields` on `MetadataStatement` / `AuthenticatorGetInfo` wholesale as a follow-up (which would preempt the next round of this), happy to open a separate PR for discussion.

Downstream: I'm pulling this into a project that verifies FIDO MDS for a government identity service (civid, https://github.com/masanork/civid) and will carry a `[patch.crates-io]` on this branch until it lands upstream.

---
🤖 Patch developed with [Claude Code](https://claude.com/claude-code) assistance.


Firstyear:
> 🤖 Patch developed with [Claude Code](https://claude.com/claude-code) assistance.

Sorry, we can't accept any AI generated contributions. 

Please consider reworking the PR without the use of AI so that we can proceed to review it.

PR #552, opened 2026-04-29
webauthn-authenticator-rs: Remove some `trace!` logs

micolous:
## Change summary

`trace` logs are normally disabled by default. This change removes a bunch of `trace` logs in `webauthn-authenticator-rs` that either:

- are commented out
- are noisy and/or limited utility
- directly expose key material (which is both noisy and of limited utility)

I've left in `trace` logs for decrypted forms of encrypted messages, because they're very useful for debugging.

While this smells security-sensitive, if an attacker _could_ enable `trace` logs and cause them to be logged somewhere they could read, then there are probably many other things that an attacker could do to recover the keys or decrypt data.

## Checklist

### DIFF
--- a/webauthn-authenticator-rs/src/cable/discovery.rs
+++ b/webauthn-authenticator-rs/src/cable/discovery.rs
@@ -301,7 +301,6 @@ impl Eid {
             }
         };
 
-        trace!("Decrypting {:?} with key {:?}", advert, key);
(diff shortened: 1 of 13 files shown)


Firstyear (approved):
Need to do a similar PR to 6.0 branch as well then.

micolous:
Looks like some of these changes are already on the 6.0 branch, but I'll reconcile that.

PR #554, opened 2026-05-07
Move `WebauthnCore::origins_match` to `webauthn-rs-proto`, don't build `tide` docs in CI

micolous:
## Change summary

This fixes a couple of build issues that have recently popped up in CI.

### Move `origins_match()` to `webauthn-rs-proto`

The fix for https://github.com/kanidm/webauthn-rs/security/advisories/GHSA-22w3-693w-x895 in `webauthn-authenticator-rs` made `webauthn-rs-core` a required dependency. However, [this was still marked as an optional dependency](https://github.com/kanidm/webauthn-rs/blob/d2c10d53ca5ef033d37ee6462e936e9eb72ad98c/webauthn-authenticator-rs/Cargo.toml#L76), breaking the build when the `crypto` feature was not enabled.

`webauthn-rs-core` was made a non-optional dependency on the 6.0 branch (https://github.com/kanidm/webauthn-rs/commit/f52f315f756f4c181526f1821534cd95b767869b), but this can't be applied to the 5.x branch because `webauthn-rs-core` _always_ depends on OpenSSL, and some `webauthn-authenticator-rs` configurations do not (eg: Windows 10 WebAuthn API).

Nothing about the fix for that bug actually requires cryptography libraries, so this PR resolves the issue by moving `WebauthnCore::origins_match()` to `webauthn-rs-proto::origin::origins_match`, which is an existing dependency of both `webauthn-rs-core` and `webauthn-authenticator-rs`.

I've put the module behind an `origin` feature flag, so that it isn't included unless actually needed for size-sensitive targets (like WASM), even though it should be optimised out automatically by the compiler.

This also fixes a couple of clippy lints.

A similar change will need to be applied to the 6.0 branch.

### Don't build docs in CI for `tide` related targets

`tide` has a transitive dependency on `rustix@0.37.28` (old version), which doesn't build on Linux with Rust nightly due to it depending on now-removed compiler internals: https://github.com/bytecodealliance/rustix/issues/1620.

I don't think there's a lot of value in building docs of the tutorials in CI anyway.

There are many other paths to _newer_ versions of `rustix` in this codebase, but these don't seem to be impacted by that issue.

## Checklist

### DIFF
--- a/.github/doc_manifest
+++ b/.github/doc_manifest
@@ -18,5 +18,4 @@ webauthn_rs/index.html
 webauthn_rs/interface/index.html
 
 webauthn_rs_core/index.html
-webauthn_rs_demo/index.html
 webauthn_rs_proto/index.html
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -254,7 +254,7 @@ jobs:
       - name: generate docs
         run: |
           cargo ${{ matrix.rust_version == 'nightly' && '+nightly' || '' }} doc --all \
-            --exclude fido-key-manager webauthn-rs-demo \
+            --exclude fido-key-manager,webauthn-rs-demo \
             --no-deps \
             --document-private-items
         env:
@@ -275,6 +275,7 @@ jobs:
       - run: |
           cargo ${{ matrix.rust_version == 'nightly' && '+nightly' || '' }} doc \
           --all \
+          --exclude webauthn-rs-demo \
           --no-deps \
           --document-private-items \
           --all-features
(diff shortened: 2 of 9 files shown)


micolous:
Ah, this PR breaks the build on Windows, because it makes OpenSSL a non-optional dependency (via `webauthn-rs-core`). This isn't a problem with the 6.0 branch because that doesn't depend on OpenSSL.

I've got another idea...

micolous:
New idea was to move the function. Also disabled some of the doc builds due to an issue with `rustix`, but I've written up another issue for solving that properly: https://github.com/kanidm/webauthn-rs/issues/555

Firstyear (approved):

PR #581, opened 2026-08-11
Credential ID Length Limited to 4096 Bytes

### LINKED ISSUE #580: Credential ID Length Enforcement
### I did this
Attempted to register a WebAuthn credential with a 1024 byte credentialId.
I was using a standard Passkey flow:
1. start_passkey_registration()
2. Create a WebAuthn Credential using the soft Authenticator through the webauthn_authenticator_rs with a 1024-byte credentialId
3. finish_passkey_registration()

### I expected the following
webauthn-rs to reject the passkey credential registration because the credentialId was too long.

[Section 6.5.1 of the WebAuthn 3 spec](https://www.w3.org/TR/webauthn-3/#sctn-attested-credential-data) specifies that credentialIdLength: "Byte length L of credentialId, 16-bit unsigned big-endian integer. Value MUST be ≤ 1023."

[Step 25 of Section 7.1 of the WebAuthn3 spec](https://www.w3.org/TR/webauthn-3/#sctn-registering-a-new-credential) specifies that "Verify that the [credentialId](https://www.w3.org/TR/webauthn-3/#authdata-attestedcredentialdata-credentialid) is ≤ 1023 bytes. Credential IDs larger than this many bytes SHOULD cause the RP to fail this [registration ceremony](https://www.w3.org/TR/webauthn-3/#registration-ceremony)."

Although the Relying Party spec says "SHOULD" instead of "MUST", I would prefer having the check in place as extremely large credential IDs could impact database storage, lookup, and payload sizes.

### What actually happened
The passkey credential was accepted and registration completed successfully. No warning or error emitted.

### Version (and git commit)
webauthn-rs version 0.5.5 off of the master branch. 
Git commit: be696b79800bd1953df78e87d0215571733cc26f

MacBook Pro M2 on MacOS Tahoe 26.5.2

### Any other comments
I discovered this issue after running the PASSKEYS-ATTACKER test suite provided by _The State of Passkeys: Studying the Adoption and Security of Passkeys on the Web_ https://www.usenix.org/system/files/conference/usenixsecurity26/sec26_prepub_jannett.pdf . This was the only Relying Party validation that I observed failing.

I have a small patch that implements the check and would be happy to submit upstream.

KozyCrypto:
## Change summary
- Adds a maximum length check for the AttestedCredentialData parser (Credentials should not have credential IDs longer than 4096 bytes)
- Adds simple internals.rs test for 4096-byte and 4097-byte credential IDs which should pass and fail, respectively

Fixes #580 

## Checklist

### DIFF
--- a/webauthn-rs-core/src/constants.rs
+++ b/webauthn-rs-core/src/constants.rs
@@ -1,2 +1,5 @@
 // Can this ever change?
 pub const CHALLENGE_SIZE_BYTES: usize = 32;
+
+// WebAuthn L3 §6.5.1 (Attested Credential Data): credentialIdLength "Value MUST be <= 1023"
+pub const CREDENTIAL_ID_MAX_LENGTH: u16 = 1023;
--- a/webauthn-rs-core/src/constants.rs
+++ b/webauthn-rs-core/src/constants.rs
@@ -2,4 +2,6 @@
 pub const CHALLENGE_SIZE_BYTES: usize = 32;
 
 // WebAuthn L3 §6.5.1 (Attested Credential Data): credentialIdLength "Value MUST be <= 1023"
-pub const CREDENTIAL_ID_MAX_LENGTH: u16 = 1023;
+// However, for compatibility with future PQC key-wrapped-keys, we will use a softer limit
+// of 4096 bytes to fit up to ML-DSA-65
+pub const CREDENTIAL_ID_MAX_LENGTH: u16 = 4096;
--- a/webauthn-rs-core/src/internals.rs
+++ b/webauthn-rs-core/src/internals.rs
@@ -1381,14 +1381,14 @@ mod tests {
         let _ = tracing_subscriber::fmt::try_init();
 
         let mut valid_acd = vec![0; 16]; // aaguid
-        valid_acd.extend_from_slice(&1023_u16.to_be_bytes());
-        valid_acd.extend_from_slice(&[0; 1023]);
+        valid_acd.extend_from_slice(&4096_u16.to_be_bytes());
+        valid_acd.extend_from_slice(&[0; 4096]);
         valid_acd.push(0xa0); // empty cbor map for public key
         assert!(super::acd_parser(&valid_acd).is_ok());
 
         let mut invalid_acd = vec![0; 16]; // aaguid
-        invalid_acd.extend_from_slice(&1024_u16.to_be_bytes());
-        invalid_acd.extend_from_slice(&[0; 1024]);
+        invalid_acd.extend_from_slice(&4097_u16.to_be_bytes());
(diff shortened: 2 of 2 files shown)


Firstyear:
Hey there,

This PR needs to be merged to https://github.com/kanidm/webauthn-rs/tree/6.0-dev-drop-openssl instead, not master as that's where we are staging 6.0. 

Second, we have to take into consideration PQC. Currently it's being discussed https://github.com/w3c/webauthn/issues/2417 and given that a Credential ID for many devices is a key-wrapped-key we have to assume that we need to fit in one of these keys. 

So I think we start with 4096 bytes as the limit which fits up to ML-DSA-65 and then if needed in future for ML-DSA-87 we can grow that number. What do you think?

KozyCrypto:
I appreciate the feedback!

> This PR needs to be merged to https://github.com/kanidm/webauthn-rs/tree/6.0-dev-drop-openssl instead, not master as that's where we are staging 6.0.

Done.

> Second, we have to take into consideration PQC. Currently it's being discussed [w3c/webauthn#2417](https://github.com/w3c/webauthn/issues/2417) and given that a Credential ID for many devices is a key-wrapped-key we have to assume that we need to fit in one of these keys.
> 
> So I think we start with 4096 bytes as the limit which fits up to ML-DSA-65 and then if needed in future for ML-DSA-87 we can grow that number. What do you think?

This is a really good point to consider.

For a hybrid PQC scenario, 4096 bytes will fit ML-DSA-65 + ECDSA P256 + AES GCM:

- ML-DSA-65 private key (4032 bytes)
- ECDSA P256 private key (32 bytes)
- AES GCM Nonce/tag (28 bytes)
- TOTAL: 4092 bytes (4 bytes to spare)

This will not fit ML-DSA-65 + ECDSA P384 or ML-DSA-87. The hybrid PQC scenario I listed above also has only 4 bytes for extra information such as rp_id or encoding. One larger raw private key size I would consider is ML-DSA-87 + ECDSA P521 + AES GCM Nonce/tag at 4990 bytes. 

Otherwise, ML-DSA private keys are large, but there are ways to deterministically expand it from a 32-byte seed, although some FIDO vendors may prefer to just use the full private key.

My ultimate goal is to implement some maximum limit for credentialIdLength. For PQC transition, I believe the reasonable limit is 4096 bytes and an absolute maximum would be 8192 bytes. 

My recommendation and current PR align with your perspective: let's limit to 4096 bytes and we can easily update as the W3C PQC work develops.

Firstyear (approved):

Firstyear:
Too easy, thank you so much for this PR!

Firstyear:
@micolous I'll leave you to have a look too

micolous (approved):

micolous:
Windows CI failures appear to be unrelated to this change.

